PR #2, opened 2022-05-06
Add Builder API schema

lightclient:
This PR adds the schema for the Builder API.


types/registration.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - description: "BLS public key of validator."
+
+SignedValidatorRegistration:
+  type: object
+  description: "The [`SignedValidatorRegistration`]() object from the Builder API spec."
+  properties:
+    message:
+      $ref: "#/ValidatorRegistration"
+    signature:
+      $ref: "../beacon-apis/types/primitive.yaml#/Signature"
```
ralexstokes, line 29: if we merge in https://github.com/ethereum/beacon-APIs/pull/209 then we can skip redefining this here

ralexstokes (commented):
looks like we lost a "status" endpoint?

builder-oapi.yaml
```diff
@@ -0,0 +1,74 @@
...
+        default: "???"
+
+tags:
+  - name: Builder
+    description: Set of endpoints to interact with an external block builder.
+paths:
+
+  /eth/v1/builder/validators:
+    $ref: "./apis/builder/validators.yaml"
+  /eth/v1/builder/{slot}/{parent_hash}/header:
```
ralexstokes, line 46: missing pubkey?

```suggestion
  /eth/v1/builder/{slot}/{parent_hash}/{pubkey}/header:
```


could also consider query params or moving this to a `POST` with the request data

apis/builder/header.yaml
```diff
@@ -0,0 +1,58 @@
...
+      description: "The validator's BLS public key."
+      schema:
+        $ref: '../../builder-oapi.yaml#/components/schemas/Pubkey'
+  responses:
+    "200":
+      description: Success response
+      content:
+        application/json:
+          schema:
+            title: ProduceBlockV2Response
```
ralexstokes, line 36:
```suggestion
            title: GetHeaderResponse
```

apis/builder/header.yaml
```diff
@@ -0,0 +1,58 @@
...
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [ phase0, altair, bellatrix ]
+                example: "phase0"
+              data:
+                oneOf:
+                 - $ref: '../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayloadHeader'
```
ralexstokes, line 45: and fields `value`, `pubkey` (of builder)

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,42 @@
...
+  summary: Submit a signed blinded block.
+  description: |
+    Submits a `SignedBlindedBeaconBlock` to the builder, binding them to the
+    block. The builder is now able to unblind the transactions in the
+    `ExecutionPayloadHeader` without the possibility of the validator modifying
+    them.
+  requestBody:
+    description: |
+      The `SignedBlindedBeaconBlock` object composed of `BlindedBeaconBlock`
+      object (produced by beacon node with `ExecutionPayloadHeader` from
```
ralexstokes, line 14:
```suggestion
      object (produced by the beacon node with `ExecutionPayloadHeader` from an
```

apis/builder/header.yaml
```diff
@@ -0,0 +1,58 @@
+get:
+  tags:
+    - Builder
+  operationId: "getHeader"
+  summary: Get an ExecutionPayloadHeader.
+  description: |
+    Requests a builder node to produce a valid execution payload header, which can integrated into a blinded beacon block and signed.
```
ralexstokes, line 7:
```suggestion
    Requests a builder node to produce a valid execution payload header, which can be integrated into a blinded beacon block and signed.
```

apis/builder/header.yaml
ralexstokes: in general, if we make these pyspec SSZ defns (see illustrative rust impl here: https://github.com/ralexstokes/mev-boost-rs/blob/d56d9c9270d3e56076fd6a8232624e4d8c8cc295/src/types.rs#L11-L29)

then we can transparently switch over to a SSZ encoding in the future rather than a (slower) JSON encoding

apis/builder/validators.yaml
```diff
@@ -0,0 +1,32 @@
...
+  tags:
+    - Builder
+  operationId: "registerValidator"
+  summary: Register or update a validator's block building preferences.
+  description: |
+    Registers a validator's preferred fee recipient and gas limit target.
+  requestBody:
+    description: |
+      The `SignedBeaconBlock` object composed of `BeaconBlock` object (produced
+      by beacon node) and validator signature.
```
ralexstokes, line 11:
```suggestion
      The `SignedValidatorRegistration` object composed of the `ValidatorRegistration` object (produced
      by the beacon node) and validator signature.
```

metachris:
Live version for this spec: https://builderspec-mevboost.metachris.com/#/Builder

types/registration.yaml
lightclient: Ooo good point, I will be sure to port changes over there.

apis/builder/header.yaml
lightclient: added in `specs/README.md`

builder-oapi.yaml
lightclient: `pubkey` is currently listed as a query parameter. I don't have a good feel for the proper formatting for this method. Originally, it felt like it was more aligned with `GET`, but after more contemplation `PUT` feels a bit more appropriate because the request _might_ cause the builder to generate a response on a different `parent_hash`. I think `PUT` is better than `POST` here because this method is mostly idempotent, because the act of calling the method does not affect the result. However, it is possible for the result to change because the builder may find a better block between requests. Since that is completely separate from the request, I think it's okay though.

apis/builder/validators.yaml
```diff
@@ -0,0 +1,37 @@
...
+  operationId: "registerValidator"
+  summary: Register or update a validator's block building preferences.
+  description: |
+    Registers a validator's preferred fee recipient and gas limit.
+
+    Builders should verify that `pubkey` corresponds to an active or pending
+    validator, and that `signature` is valid under `pubkey`. Otherwise, builder
+    must error. Requests with `timestamp` less than or equal to the previous
+    announcement must error. Requests with `timestamp` greater than 10 seconds
+    +/- the builder's local time must error.
```
metachris, line 13:
```suggestion
    announcement must error. Requests with `timestamp` greater than 10 seconds
    in the future must error.
```
Old messages must still be valid, for instance if a new builder comes online.

apis/builder/header.yaml
```diff
@@ -0,0 +1,69 @@
...
+  operationId: "getHeader"
+  summary: Get an `ExecutionPayloadHeader`.
+  description: |
+    Requests a builder node to produce a valid execution payload header, which
+    can integrated into a blinded beacon block and signed.
+
+    Builder must immediately return the header that is able to pay
+    `feeRecipient` the most. If the builder is unaware of `parent_hash`, it
+    must error. If `pubkey` does not match the builder's expected proposer
+    `pubkey` for `slot, it must error. If the builder has not received a
```
metachris, line 13:
```suggestion
    must error. If `pubkey` does not match the builder's expected proposer
    `pubkey` for `slot`, it must error. If the builder has not received a
```

apis/builder/header.yaml
metachris: I'm not following - where are the fields for `value`, `pubkey` (of builder, and `signature` (of the relay) in this response?

builder-oapi.yaml
metachris: I think it would be best to have header first, like this:
```/eth/v1/builder/header```

Maybe put all arguments as query args, rather than some in the path and others as query args? 🤔 

I guess any method works, but imo GET seems fine because it will not modify anything on the relay. The relay will just return whatever block it has for these settings or none. This call will probably not be forwarded to the builder anyways (when we were thinking through relay+builder interaction, it seems natural for builders to stream new blocks to the relay, mostly for timing reasons).

Comment (lightclient) [reply builder-oapi.yaml:45]: I will stick with `GET` for now then, but I do think path vars are better because query parameters typically filter results whereas we need all three values to specify the resource we're requesting.

File: apis/builder/validators.yaml (hunk 8 above)
Comment (lightclient) [reply apis/builder/validators.yaml:13]: Good point - updated.

File: apis/builder/header.yaml [hunk 10]
```diff
@@ -0,0 +1,69 @@
...
+            title: GetHeaderResponse
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [ bellatrix ]
+                example: "bellatrix"
+              data:
+                oneOf:
+                 - $ref: "../../builder-oapi.yaml#/components/schemas/SignedBuilderBid"
```
metachris, line 56:
```suggestion
              data:
                type: object
                properties:
                 - $ref: "../../builder-oapi.yaml#/components/schemas/SignedBuilderBid"
```

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,42 @@
+post:
+  operationId: "submitBlindedBlock"
+  summary: Submit a signed blinded block.
```
metachris, line 3:
```suggestion
  summary: Submit a signed blinded block and receive an execution payload.
```

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - Builder
+  requestBody:
+    description: |
+      A `SignedBlindedBeaconBlock`.
+    required: true
+    content:
+      application/json:
+        schema:
+          oneOf:
+           - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
```
metachris, line 22:
```suggestion
        schema:
          type: object
          properties:
           - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
```

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,42 @@
...
+  responses:
+    "200":
+      description: Success response.
+      content:
+        application/json:
+          schema:
+            oneOf:
+              - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
```
metachris, line 30:
```suggestion
          schema:
            type: object
            properties:
              - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
```

types/bid.yaml
```diff
@@ -0,0 +1,22 @@
...
+      description: "BLS public key of builder."
+
+SignedBuilderBid:
+  type: object
+  description: "The `SignedBuilderBid` object from the Builder API spec."
+  properties:
+    message:
+      $ref: "#/BuilderBid"
+    signature:
+      $ref: "../beacon-apis/types/primitive.yaml#/Signature"
```
metachris, line 22:
```suggestion
      $ref: "../beacon-apis/types/primitive.yaml#/Signature"
      description: "BLS signature of the relay"      
```

Clarify that this is the signed by the relay?

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,49 @@
...
+            title: SubmitBlindedBlockResponse
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [ bellatrix ]
+                example: "bellatrix"
+              data:
+                oneOf:
+                  - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
```
metachris, line 37:
```suggestion
              data:
                 $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
```

metachris:
> looks like we lost a "status" endpoint?

https://github.com/lightclient/builder-specs/pull/1/files

README.md
```diff
@@ -1,51 +1,99 @@
...
 ```
 
+### Render API Specification
+
+To render spec in browser, you will simply need an HTTP server to load the
+`index.html` file in root of the repo.
+
+For example:
+```console
+python -m http.server 8080
```
metachris, line 64:
```suggestion
python3 -m http.server 8080
```

builder-oapi.yaml
```diff
@@ -0,0 +1,74 @@
...
+  license:
+    name: "CC0-1.0"
+    url: "https://creativecommons.org/publicdomain/zero/1.0/"
+
+servers:
+  - url: "{server_url}"
+    variables:
+      server_url:
+        description: "Builder node API url"
+        default: "???"
```
metachris, line 39:
```suggestion
        default: ""
```

types/bid.yaml
lightclient: There isn't a concept of a "relay" in this API though, the only specified entities are the CL client and the external builder.

README.md
lightclient: As I mentioned [here](https://github.com/ethereum/beacon-APIs/pull/213, most people are running python 3 now. I think it is unnecessary to additional decoration.

apis/builder/blinded_blocks.yaml
lightclient: I think it's better to leave these. Eventually we're going to need to support multiple object types here, so we might as well come up with a solution that also works for `oneOf`. Unfortunately, it appears this is a [longstanding issue](https://github.com/swagger-api/swagger-ui/issues/3803) with the swagger ui. I think the best work around is to have statically defined examples. I've implemented that in 7d21dc0. In the future, it might be nice to have a valid example value (e.g. the signature checks out).

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,56 @@
...
+    description: A `SignedBlindedBeaconBlock`.
+    required: true
+    content:
+      application/json:
+        schema:
+          oneOf:
+            - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
+        examples:
+          bellatrix:
+            $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBlindedBeaconBlock"
```
metachris, line 24:
```suggestion
        schema:
          $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
```
simplify?

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,56 @@
...
+              version:
+                type: string
+                enum: [ bellatrix ]
+                example: "bellatrix"
+              data:
+                oneOf:
+                  - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
+          examples:
+            bellatrix:
+              $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.ExecutionPayload"
```
metachris, line 44:
```suggestion
              data:
                $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
```
simplify?

apis/builder/blinded_blocks.yaml
lightclient: These are different components.

apis/builder/header.yaml
```diff
@@ -0,0 +1,72 @@
...
+          examples:
+            bellatrix:
+              $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBuilderBid"
```
metachris, line 59: Why are you using a hardcoded example? If you remove the `oneOf` the example is automatic, and when updating we wouldn't need to remember to also update the example json.

apis/builder/blinded_blocks.yaml
metachris: How do you see differing objects in the requests/responses? If that's anticipated I'm fine with using `oneOf`, although I'm not sure what those future cases might be. 

I tend towards keeping things simple in the beginning and only adding complexity once actually needed.

README.md
```diff
@@ -1,51 +1,99 @@
...
+tampering with them.
+
+This design is based on the original proposal for trusted external builders:
+["MEV-Boost: Merge ready Flashbots Architecture"](mev-boost-ethr).
 
-Local changes will be observable if "dev" is selected in the "Select a definition" drop-down in the web UI.
+#### Builder software
 
-Users may need to tick the "Disable Cache" box in their browser's developer tools to see changes after modifying the source. 
+Users will typically connect thier CL clients to builders with builder
```
tersec, line 38: their

apis/builder/blinded_blocks.yaml
```diff
@@ -0,0 +1,56 @@
+post:
+  operationId: "submitBlindedBlock"
+  summary: Submit a signed blinded block and get unblinded execution payload.
+  description: |
+    Submits a `SignedBlindedBeaconBlock` to the builder, binding the proposer
```
tersec, line 5: There doesn't appear to be any SSZ definition of a `SignedBlindedBeaconBlock` in `specs/README.md`.

apis/builder/header.yaml
```diff
@@ -0,0 +1,72 @@
+get:
+  operationId: "getHeader"
+  summary: Get an execution payload header.
+  description: |
+    Requests a builder node to produce a valid execution payload header, which
+    can integrated into a blinded beacon block and signed.
```
rootulp, line 6:
```suggestion
    can be integrated into a blinded beacon block and signed.
```

builder-oapi.yaml
```diff
@@ -0,0 +1,84 @@
+openapi: "3.0.3"
+
+info:
+  title: Builder API
+  description: |
+    API specification for external builder nodes. This interface enables
+    validators to delegate block building duties, while allowing for
```
rootulp, line 7:
```suggestion
    validators to delegate block building duties, while allows for
```

apis/builder/blinded_blocks.yaml
lightclient: Thanks, I've updated it to include `SignedBlindedBeaconBlock` and `SignedBuilderBidV1`.

specs/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Some objects can be updated independently of the `consensus-specs`, because
+they originate soley from this specification. The objects are postfixed with
+`VX` to denote their revision.
+
+##### `ValidatorRegistrationV1`
+
+```python
+class ValidatorRegistrationV1(Container):
+    feeRecipient: Bytes20
```
tersec, line 23: `feeRecipient` and `gasLimit` use an inconsistent case-convention with the rest of these SSZ objects (`proposer_index`, `parent_root`, `state_root`, `randao_reveal`, et cetera).

specs/README.md
```diff
@@ -0,0 +1,119 @@
...
+    gasLimit: uint64
```
metachris, line 24:
```suggestion
    fee_recipient: Bytes20
    gas_limit: uint64
```

builder-oapi.yaml
lightclient: fixed

examples/bellatrix/signed_blinded_beacon_block.json
```diff
@@ -0,0 +1,147 @@
...
+      "parent_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "state_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "body": {
+        "randao_reveal": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505",
+        "eth1_data": {
+          "deposit_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+          "deposit_count": "1",
+          "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
+        },
+        "graffiti": "string",
```
ralexstokes, line 15: I think this is technically hex-encoded
```suggestion
        "graffiti": "0xdeadbeefcafe",
```

types/bellatrix/bid.yaml
```diff
@@ -0,0 +1,23 @@
...
+  BuilderBid:
+    type: object
+    description: "The `BuilderBid` object from the Builder API spec."
+    properties:
+      header:
+        $ref: "../../beacon-apis/types/bellatrix/execution_payload.yaml#/Bellatrix/ExecutionPayloadHeader"
+        description: "`ExecutionPayloadHeader` to use in block proposal."
+      value:
+        $ref: "../../beacon-apis/types/primitive.yaml#/Uint64"
+        description: "Payment in wei that will be paid to the `fee_recipient` account."
```
ralexstokes, line 11: if this is a `wei` amount then we want `Uint256` here:

```suggestion
        $ref: "../../beacon-apis/types/primitive.yaml#/Uint256"
        description: "Payment in wei that will be paid to the `fee_recipient` account."
```

specs/README.md
```diff
@@ -0,0 +1,136 @@
...
+## Containers
+
+Consider the following definitions supplementary to the definitions in
+[`consensus-specs`][consensus-specs]. For information on how containers are
+signed, see [Signing](#signing).
+
+### Independently Versioned
+
+Some objects can be updated independently of the `consensus-specs`, because
+they originate soley from this specification. The objects are postfixed with
```
ralexstokes, line 28:
```suggestion
they originate solely from this specification. The objects are postfixed with
```

specs/README.md
```diff
@@ -0,0 +1,136 @@
...
+`VX` to denote their revision.
+
+#### `ValidatorRegistrationV1`
+
+```python
+class ValidatorRegistrationV1(Container):
+    fee_recipient: Bytes20
```
ralexstokes, line 35: can use this definition from the bellatrix spec if you prefer

```suggestion
    fee_recipient: ExecutionAddress
```

apis/builder/header.yaml
```diff
@@ -0,0 +1,72 @@
...
+    Requests a builder node to produce a valid execution payload header, which
+    can be integrated into a blinded beacon block and signed.
+
+    Builder must immediately return the header that is able to pay
+    `feeRecipient` the most. If the builder is unaware of `parent_hash`, it
+    must error. If `pubkey` does not match the builder's expected proposer
+    `pubkey` for `slot`, it must error. If the builder has not received a
+    validator registration associated with `pubkey`, it must error.
+
+    When possibe, the builder must return a header with `gas_limit` equal to
```
ralexstokes, line 14:
```suggestion
    When possible, the builder must return a header with `gas_limit` equal to
```

ralexstokes (commented):
left a few minor things to address but starting to look really good!!

types/bellatrix/bid.yaml
lightclient: Good catch.

examples/bellatrix/signed_builder_bid.json
```diff
@@ -0,0 +1,28 @@
+{
+  "value": {
```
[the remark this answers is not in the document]
lightclient: yeah this is just some weirdness with the UI - if I define it outside this file there is a weird `$$ref: asdfasdf` object at the end.

apis/builder/validators.yaml
```diff
@@ -0,0 +1,37 @@
+post:
+  operationId: "registerValidator"
+  summary: Register or update a validator's block building preferences.
+  description: |
+    Registers a validator's preferred fee recipient and gas limit.
+
+    Builders should verify that `pubkey` corresponds to an active or pending
+    validator, and that `signature` is valid under `pubkey`. Otherwise, builder
+    must error. Requests with `timestamp` less than or equal to the previous
+    announcement must error. Requests with `timestamp` more than 10 seconds in
```
djrtwo, line 10: making sure builders don't track timestamps of "error" calls (both bad sigs and too future of timestamps)
```suggestion
    successful announcement must error. Requests with `timestamp` more than 10 seconds in
```

apis/builder/validators.yaml
```diff
@@ -0,0 +1,37 @@
...
+    the future must error.
+  tags:
+    - Builder
+  requestBody:
+    description: An array of signed validator block building preferences.
```
djrtwo, line 15: It's unclear to me the value of making this an array rather than a singleton especially in light of potentially sending this to gossip. In the event that this registration is gossiped, it would certainly be as singletons.

That said, mev-boost could split into singletons and gossip.

apis/builder/validators.yaml
```diff
@@ -0,0 +1,37 @@
...
+    "400":
+      description: Error response.
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: "../../builder-oapi.yaml#/components/schemas/ErrorMessage"
+              - example:
+                  code: 400
+                  message: "Unknown validator: 0x933ad9491b62059dd065b560d256d8957a8c402cc6e8d8ee7290ae11e8f7329267a8811c397529dac52ae1342ba58c95"
```
djrtwo, line 35: Because of the general pass-through of this from mev-boost to builders, it's unclear which piece of software in he line is supposed to assess the errors. Is mev-boost supposed to see "unknown val" or is this actually a builder (or relay) response?

specs/README.md
```diff
@@ -0,0 +1,136 @@
...
+All signature operations should follow the [standard BLS operations][bls]
+interface defined in `consensus-specs`.
+
+There are two types of data to sign over in the Builder API:
+* In-protocol messages, e.g. [`BlindedBeaconBlock`](#blindedbeaconblock), which
+  should compute the signing root using [`compute_signing_root`][compute-root]
+  and use the domain specified for beacon block proposals.
+* Builder API messages, e.g. validator registration, which should compute the
+  signing root using [`compute_signing_root`][compute-root] and the domain
+  `DomainType('0x10000001')`.
```
djrtwo, line 120: Suggest `0x00000001` as this is the 0th application to use this bitmask.
or maybe `0x01000001` if you like 1-indexing

but starting counting at 0 seems natural.
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#domain-types

types/registration.yaml
```diff
@@ -0,0 +1,35 @@
+ValidatorRegistration:
```
djrtwo, line 1: Should this type be versioned?

djrtwo: I know it's not in relation to a fork namespace (and shouldn't be), but it seems likely to need to be upgraded at some point

djrtwo (commented):
nice! did a pass. looking good

apis/builder/validators.yaml
ralexstokes: The structure here just mirrors the signature in https://github.com/ethereum/beacon-APIs/pull/209

and the rationale is to allow nodes with many validators to make one request, rather than make many requests one-by-one

apis/builder/validators.yaml
ralexstokes: I think it is unspecified and you would likely have checks wherever it is reasonable to do so

I intend to add a consensus API client to https://github.com/ralexstokes/mev-boost-rs so that it can access consensus info and in that case it would make sense to validate this before even passing on to builders

But you could also have a super thin `mev-boost` that just lets the relay/builder deal with it

types/registration.yaml
ralexstokes: if we version this, it should be updated in https://github.com/ethereum/beacon-APIs/pull/209

apis/builder/validators.yaml
ralexstokes: we can also more naturally get fine-grained error reporting if we move to the individual style interface

right now you can return an error from this endpoint but there doesn't seem to be a way to specify which registration in the batch failed (unless identifying info, like pubkey, is in the error message)

apis/builder/validators.yaml
lightclient: I tend to imagine `mev-boost` as a pretty thin layer, so in this case the builder/relay would go to register the validator and realize the pubkey is not in the active/pending set and error.

types/registration.yaml
lightclient: I'm going to remove this type as we can now refer to it directly in the beacon api. Is there an example of a beacon api type that has been updated? Seems like a combo of updating the endpoint version and fork-based object updates so far.

apis/builder/validators.yaml
lightclient: remove in 565d1db

apis/builder/header.yaml
```diff
@@ -0,0 +1,72 @@
+get:
+  operationId: "getHeader"
+  summary: Get an execution payload header.
+  description: |
+    Requests a builder node to produce a valid execution payload header, which
+    can be integrated into a blinded beacon block and signed.
+
+    Builder must immediately return the header that is able to pay
+    `feeRecipient` the most. If the builder is unaware of `parent_hash`, it
```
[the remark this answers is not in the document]
lightclient: done

apis/builder/header.yaml
```diff
@@ -0,0 +1,72 @@
...
+    must error. If `pubkey` does not match the builder's expected proposer
+    `pubkey` for `slot`, it must error. If the builder has not received a
```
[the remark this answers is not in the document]
lightclient: Noted in https://github.com/ethereum/builder-specs/issues/5 -- will address in the next release.

README.md
```diff
@@ -1,51 +1,99 @@
...
+```
+
+The spec will render at [http://localhost:8080](http://localhost:8080).
+
+### Usage
+
+Local changes will be observable if "dev" is selected in the "Select a
+definition" drop-down in the web UI.
+
+It may be neccessary to tick the "Disable Cache" box in their browser's
```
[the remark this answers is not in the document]
lightclient: done in 85b0c486

lightclient:
I really appreciate everyone's feedback. I'm going to go ahead and merge this as the feedback has slowed and things seems to be in a pretty solid place.

PR #3, opened 2022-05-06
Add `dist` folder

### LINKED ISSUE #1: Missing web assets
Repo seems to be missing a `/dist` dir with CSS, image, etc. assets.

If you look at the `beacon-APIs` repo, this folder is just checked in (so I don't think it can be generated locally by some tooling).

I fixed the website render by cloning this repo along w/ the beacon APIs submodule and just copying `/dist` next to `index.html`.

We should probably do the same here.

lightclient:
Fixes #1


ralexstokes (approved):
l g t m.

looks like CI will fail until we merge #2 tho

PR #15, opened 2022-05-16
clarify signing for builder messages

### LINKED ISSUE #14: Clarify signing routines
The signing routines for the Builder API messages are under-specified.

The spec currently says to compute a message to sign over using `compute_signing_root` which takes the SSZ data and a `Domain`. Rather than specify a `Domain`, the spec says to simply use the `DomainType` `0x00000001`.

There are two degrees of freedom to go from a `DomainType` to a `Domain` which we should decide how to handle:

1) Fork version
2) Genesis validators root

The basic question to answer is if we want to version messages across instances of the protocol or not. The degrees of freedom here are to allow for messages (like deposits) where the relevant information either is not known or we don't want to restrict a message to a given fork.

If we follow the current layout of the builder spec here, we **should** version (so use `state.genesis_validators_root` and the appropriate `fork_version`) the **builder bid types** but **not** version (so use "default" values as defined in the spec) the **validator registration types**. Moreover, this maps to the "deposits vs. other protocol messages" usage in the consensus specs.

ralexstokes:
fixes #14.

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -2,6 +2,8 @@
 
 ## Table of Contents
 
+* [Constants](#constants)
+    * [Domain types](#domain-types)
 * [Containers](#containers)
     * [Independently Versioned](#independentlyversioned)
         * [`ValidatorRegistrationV1`](#validatorregistrationv1)
@@ -15,6 +17,14 @@
             * [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
     * [Signing](#signing)
 * [Endpoints](#endpoints)
+
+## Constants
+
+### Domain types
+
+| Name | Value |
+| - | - |
+| `DOMAIN_APPLICATION_BUILDER` | `DomainType('0x00000001')` |
 
 ## Containers
 
@@ -111,18 +121,20 @@ class SignedBlindedBeaconBlock(Container):
 All signature operations should follow the [standard BLS operations][bls]
 interface defined in `consensus-specs`.
 
+To assist in signing, we use a function from the [consensus specs][consensus-specs]: 
+[`compute_domain`][compute-domain]
+
 There are two types of data to sign over in the Builder API:
 * In-protocol messages, e.g. [`BlindedBeaconBlock`](#blindedbeaconblock), which
   should compute the signing root using [`compute_signing_root`][compute-root]
   and use the domain specified for beacon block proposals.
 * Builder API messages, e.g. validator registration, which should compute the
-  signing root using [`compute_signing_root`][compute-root] and the domain
-  `DomainType('0x00000001')`.
+  signing root using [`compute_signing_root`][compute-root] with domain given by
+  `compute_domain(DOMAIN_APPLICATION_BUILDER)`.
 As `compute_signing_root` takes `SSZObject` as input, client software should
 convert in-protocol messages to their SSZ representation to compute the signing
 root and Builder API messages to the SSZ representations defined
 [above](#containers).
-
 
 ## Endpoints
 
@@ -134,3 +146,4 @@ https://ethereum.github.io/builder-specs/.
 [consensus-specs]: https://github.com/ethereum/consensus-specs
 [bls]: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#bls-signatures
 [compute-root]: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#compute_signing_root
+[compute-domain]: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#compute_domain


specs/README.md
```diff
@@ -16,6 +18,14 @@
     * [Signing](#signing)
 * [Endpoints](#endpoints)
 
+## Constants
+
+### Domain types
+
+| Name | Value |
+| - | - |
+| `DOMAIN_APPLICATION_BUILDER` | `DomainType('0x00000000')` |
```
mcdee, line 27: I strongly recommend that there is something set here, otherwise the domain will equal the mask and that seems a potential security risk.

Also, to date the domain types have generally been specific to operations so may make sense to make this more focused than just "builder".

specs/README.md
```diff
@@ -111,19 +121,27 @@ class SignedBlindedBeaconBlock(Container):
 All signature operations should follow the [standard BLS operations][bls]
 interface defined in `consensus-specs`.
 
+To assist in signing, we define a helper function referencing functionality
+from the [consensus specs](consensus-specs):
+
+```python
+def compute_builder_domain() -> Domain:
+    domain_type = DomainType(DOMAIN_APPLICATION_MASK & DOMAIN_APPLICATION_BUILDER)
```
mcdee, line 129: If I'm reading this correctly it ends up with a domain type of 0.  Should this be `|` rather than `&`?

specs/README.md
ralexstokes: the idea is to "index" applications which we can just start counting from 0, 1, 2, ...

but I'd be fine if we go with `0x01` to avoid the degenerate case

ralexstokes: > Also, to date the domain types have generally been specific to operations so may make sense to make this more focused than just "builder".

do you have a suggestion? not attached to this name

specs/README.md
ralexstokes: yes sorry, thanks, I'll update

specs/README.md
mcdee: Domains have previously been quite specific, _e.g._ `DOMAIN_DEPOSIT` and `DOMAIN_VOLUNTARY_EXIT`, so perhaps `DOMAIN_BUILDER_REGISTRATION` and `DOMAIN_BUILDER_BLOCK` or similar.

ralexstokes: yeah we could have two:

`DOMAIN_BUILDER_REGISTRATION`
`DOMAIN_BUILDER_BID`

lightclient: I'm a little worried about setting the precedent that if you are an application that needs a domain type, you need a domain type for _each_ of your operations.

lightclient: This also feels wrong -- `DOMAIN_APPLICATION_BUILDER` is actually not the `DomainType` for the application (contrary to the heading), but rather the index of the application in the `DOMAIN_APPLICATION_MASK` space.

lightclient (changes requested):
I feel like we either need to better define an "application index in the application mask space" or `DOMAIN_APPLICATION_BUILDER` should just be `0x00000001`. Right now it seems to be claiming `0x00000000` is the `DomainType` for the builder, but then in `compute_builder_domain` it actually bitwise-ors it with `DOMAIN_APPLICATION_MASK`.

specs/README.md
realbigsean: You need domain separation per thing you're signing so a signature for one object can't be interpreted as a valid signature for another object.

mcdee: What do you mean by "subdomain" here?

ralexstokes:
ok I've update the PR to a "lowest common denominator"

I think the only other decision we should make here is if we change the domain so there is less risk of misuse/collisions/ etc.

If we don't like the `DOMAIN_APPLICATION_BUILDER` as defined then I'd suggest this:

`DomainType(0xdd70c7b1)`

this value is the first four bytes of the hash of the python term `b"ethereum-mev-builder"` with the top bit set.

we should separately consider adding versioning to the builder bids but if there is support to merge this, we can cut a release and give everyone something stable to start testing on kiln

realbigsean:
I actually think `DOMAIN_APPLICATION_BUILDER` needs to be spec'd in terms of the bit mask because it's configurable in the consensus spec and if it changes, this breaks

ralexstokes:
@realbigsean 


>  it's configurable in the consensus spec and if it changes, this breaks

it shouldn't change tho -- from the spec:

> The following values are (non-configurable) constants used throughout the specification.

we can be explicit about the bit mask but it seems to be causing some confusion; in lieu of taking more time to sort that out I was going to just assert there is some `DomainType` -- it has some extra semantics wrt the consensus spec but the builder specs here don't care about that

realbigsean:
> @realbigsean
> 
> > it's configurable in the consensus spec and if it changes, this breaks
> 
> it shouldn't change tho -- from the spec:
> 
> > The following values are (non-configurable) constants used throughout the specification.

Ah true, this is good with me then.

specs/README.md
lightclient: @mcdee an application could define a wrapper for `compute_signing_root` like this:

```python
def compute_app_signing_root(ssz_object: SSZObject, subdomain: Domain) -> Root:
    return eth2spec.mainnet.spec.compute_signing_root(SigningData(
        object_root=hash_tree_root(ssz_object)
        domain=subdomain,
    ), compute_domain(DOMAIN_TYPE_APPLICATION_XYZ))
``` 

This way, the outer object is in the domain of the application and the inner object has an application-specific subdomain.

PR #18, opened 2022-05-18
Add validator spec

### LINKED ISSUE #13: Clarify validator responsibilities
Validators are expected to periodically (once per epoch? once per N epochs?) submit a `SignedValidatorRegistration` to builder software but I do not think this is document anywhere outside of the comment in the https://github.com/ethereum/beacon-APIs

We should craft a `spec/validator.md` that contains some light prose on the expected validator flow to be able to participate in the external builder network.

ralexstokes:
Fixes #13.

### DIFF
--- a/specs/builder.md
+++ b/specs/builder.md
@@ -1,22 +1,27 @@
-# Builder API
+# Builder spec
 
 ## Table of Contents
 
-* [Constants](#constants)
-    * [Domain types](#domain-types)
-* [Containers](#containers)
-    * [Independently Versioned](#independentlyversioned)
-        * [`ValidatorRegistrationV1`](#validatorregistrationv1)
-        * [`SignedValidatorRegistrationV1`](#signedvalidatorregistrationv1)
-    * [Fork Versioned](#forkversioned)
-        * [Bellatrix](#forkversioned)
-            * [`BuilderBid`](#builderbid)
-            * [`SignedBuilderBid`](#signedbuilderbid)
-            * [`BlindedBeaconBlockBody`](#blindedbeaconblockbody)
-            * [`BlindedBeaconBlock`](#blindedbeaconblock)
-            * [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
-    * [Signing](#signing)
-* [Endpoints](#endpoints)
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Constants](#constants)
+  - [Domain types](#domain-types)
+- [Containers](#containers)
+  - [Independently Versioned](#independently-versioned)
+    - [`ValidatorRegistrationV1`](#validatorregistrationv1)
+    - [`SignedValidatorRegistrationV1`](#signedvalidatorregistrationv1)
+  - [Fork Versioned](#fork-versioned)
+    - [Bellatrix](#bellatrix)
+      - [`BuilderBid`](#builderbid)
+      - [`SignedBuilderBid`](#signedbuilderbid)
+      - [`BlindedBeaconBlockBody`](#blindedbeaconblockbody)
+      - [`BlindedBeaconBlock`](#blindedbeaconblock)
+      - [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
+  - [Signing](#signing)
+- [Endpoints](#endpoints)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Constants
 
--- a/specs/builder.md
+++ b/specs/builder.md
@@ -1,4 +1,4 @@
-# Builder spec
+# Builder Specification
 
 ## Table of Contents
 
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -31,7 +31,7 @@ participate in an external builder network.
 
 At a high-level, there is a registration step validators must perform ahead of any proposal duties so builders know how
 to craft blocks for their specific proposal. Having performed the registration, a validator waits until it is their turn
-to propose the next block in the chain. The validator then requests an `ExecutionPayload` from the external builder 
+to propose the next block in the chain. The validator then requests an `ExecutionPayload` from the external builder
 network to put into their `SignedBeaconBlock` in lieu of one they could build locally.
 
 ## Prerequisites
@@ -53,7 +53,7 @@ builders it wants to utilize during block production.
 
 ### Preparing a registration
 
-To do this, the validator client assembles a [`ValidatorRegistrationV1`][validator-registration-v1] with the following 
+To do this, the validator client assembles a [`ValidatorRegistrationV1`][validator-registration-v1] with the following
 information:
 
 * `fee_recipient`: an execution layer address where fees for the validator should go.
@@ -68,7 +68,7 @@ The validator takes the constructed `ValidatorRegistrationV1` `message` and sign
 the [Builder spec][builder-spec] to make a `signature`.
 
 This `signature` is placed along with the `message` into a `SignedValidatorRegistrationV1` and submitted to a connected
-beacon node using the [`registerValidator`][register-validator-api] endpoint of the standard validator 
+beacon node using the [`registerValidator`][register-validator-api] endpoint of the standard validator
 [beacon node APIs][beacon-node-apis].
 
 Validators **should** submit valid registrations well ahead of any potential beacon chain proposal duties to ensure
@@ -77,13 +77,13 @@ their building preferences are widely available in the external builder network.
 ### Registration dissemination
 
 Validators are expected to periodically send their own `SignedValidatorRegistrationV1` messages upstream to the external
-builder network using the [`registerValidator`][register-validator-with-builder] endpoint of the standard 
+builder network using the [`registerValidator`][register-validator-with-builder] endpoint of the standard
 [APIs defined in the builder spec][builder-spec-apis].
 
 Registrations should be re-submitted frequently enough that any upstream builder software that comes online is aware of
 each validator's latest registration in a timely manner.
 
-This specification suggests validators re-submit to builder software every 
+This specification suggests validators re-submit to builder software every
 `EPOCHS_PER_VALIDATOR_REGISTRATION_SUBMISSION` epochs.
 
 ## Beacon chain responsibilities
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -122,8 +122,8 @@ external builder network fails to provide a `SignedBuilderBid` in time in order 
 `SignedBeaconBlock` during the proposer's slot.
 
 Honest proposers using the external builder network will give the builders a duration of
-`BUILDER_PROPOSAL_DELAY_TOLERANCE` before the external builder is considered to have hit the deadline and the external
-builder flow must be aborted in favor of a local build process.
+`BUILDER_PROPOSAL_DELAY_TOLERANCE` to provide a `SignedBuilderBid` before the external builder is considered to have hit
+the deadline and the external builder flow must be aborted in favor of a local build process.
 
 **WARNING**: Validators must be careful to not get slashed when orchestrating the duplicate building pathways.
   See the [section on slashing](#proposer-slashing) for more information.
@@ -141,7 +141,10 @@ To avoid slashing when using the external builder network, a validator should be
 parameters to do so. Regardless of which process completes in time, the validator should cancel the other
 process as soon as they have produced a signature for a beacon block, either a `BeaconBlock` **or** a
 `BlindedBeaconBlock`. Producing distinct signatures for the validator's proposal slot, for example because the
- transactions list of the `BeaconBlockBody` are different, is the slashable offense.
+transactions list of the `BeaconBlockBody` are different, is the slashable offense. This means if a validator publishes
+a signature for a `BlindedBeaconBlock` (via a dissemination of a `SignedBlindedBeaconBlock`) then the validator
+**MUST** not use the local build process as a fallback, even in the event of some failure with the external builder
+network.
 
 [builder-spec]: ./builder.md
 [builder-spec-apis]: ./builder.md#endpoints
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -146,6 +146,13 @@ a signature for a `BlindedBeaconBlock` (via a dissemination of a `SignedBlindedB
 **MUST** not use the local build process as a fallback, even in the event of some failure with the external builder
 network.
 
+## Responsibilites during the Merge transition
+
+Honest validators will not utilize the external builder network until after the transition from the proof-of-work chain
+to the proof-of-stake beacon chain has been finalized by the proof-of-stake validators. This requirement is in place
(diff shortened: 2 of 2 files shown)


specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+This `signature` is placed along with the `message` into a `SignedValidatorRegistrationV1` and submitted to a connected
+beacon node using the [`registerValidator`][register-validator-api] endpoint of the standard validator 
+[beacon node APIs][beacon-node-apis].
+
+Validators **should** submit valid registrations well ahead of any potential beacon chain proposal duties to ensure
+their building preferences are widely available in the external builder network.
+
+### Registration dissemination
+
+Beacon nodes are expected to keep track of the latest (determined by `timestamp`) registrations for each validator it
```
g11tech, line 76: can we move this expectation to validator and let the beacon be a thin layer to pass on this information to the relay, adding a duty to a validator which can also sign latest timestamp looks to be an ideal place. BNs can just track this for their own prepare payload calls to the engine (in parallel)

metachris: Previously my impression was resubmitting is the job of the validator and mev-boost, not the BN.

realbigsean: Agree registration re-submission should be the VC responsibility rather than the BN, and the BN act mainly as a proxy.  Since validator registration is independent of changes to the head of the BN (so requires no info from the BN) it's different from the other "proposer preparation" responsibilities which do more or less necessitate the BN playing a role in re-submission.

realbigsean: By "other proposer preparation responsibilities" I really mean local payload attribute submission

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+   upstream builder software.
+4. The upstream builder software in turn responds with the full `ExecutionPayload`. The proposer can use this payload
+   to assemble a `SignedBeaconBlock` following the rest of the proposal process outlined in the
+   [Bellatrix specs][bellatrix-specs].
+
+#### Relation to local block building
+
+The external builder network offers a service for proposers that may from time to time fail to produce a timely block.
+Honest proposers who elect to use the external builder network **MUST** also build a block locally in the event the
+external builder network fails to provide the full `ExecutionPayload` in time to propagate the full block during the
```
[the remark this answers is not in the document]
metachris: Very much agree. Should be bold and mention that the result of still proposing would likely be slashing.

specs/builder.md
```diff
@@ -1,20 +1,25 @@
-# Builder API
+# Builder spec
```
lightclient, line 1:
```suggestion
# Builder Specification
```

specs/builder.md
```diff
@@ -1,20 +1,25 @@
...
-    * [Fork Versioned](#forkversioned)
-        * [Bellatrix](#forkversioned)
-            * [`BuilderBid`](#builderbid)
-            * [`SignedBuilderBid`](#signedbuilderbid)
-            * [`BlindedBeaconBlockBody`](#blindedbeaconblockbody)
-            * [`BlindedBeaconBlock`](#blindedbeaconblock)
-            * [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
-    * [Signing](#signing)
-* [Endpoints](#endpoints)
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
lightclient, line 5: epic

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Introduction
+
+This document explains the way in which a beacon chain validator is expected to use the [Builder spec][builder-spec] to
+participate in an external builder network.
+
+At a high-level, there is a registration step validators must perform ahead of any proposal duties so builders know how
+to craft blocks for their specific proposal. Having performed the registration, a validator waits until it is their turn
+to propose the next block in the chain. The validator then requests an `ExecutionPayload` from the external builder
```
lightclient, line 32:
```suggestion
to propose the next block in the chain. The validator then requests an `ExecutionPayloadHeader` from the external builder 
```

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Introduction
+
+This document explains the way in which a beacon chain validator is expected to use the [Builder spec][builder-spec] to
+participate in an external builder network.
+
+At a high-level, there is a registration step validators must perform ahead of any proposal duties so builders know how
+to craft blocks for their specific proposal. Having performed the registration, a validator waits until it is their turn
+to propose the next block in the chain. The validator then requests an `ExecutionPayload` from the external builder 
+network to put into their `SignedBeaconBlock` in lieu of one they could build locally.
```
lightclient, line 33:
```suggestion
network to put into their `SignedBlindedBeaconBlock` in lieu of one they could build locally.
```

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Prerequisites
+
+This document assumes knowledge of the terminology, definitions and other material in the [Builder spec][builder-spec]
```
lightclient, line 37:
```suggestion
This document assumes knowledge of the terminology, definitions, and other material in the [Builder spec][builder-spec]
```
feels-bad-man.jpg

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+builders it wants to utilize during block production.
+
+### Preparing a registration
+
+To do this, the validator client assembles a [`ValidatorRegistrationV1`][validator-registration-v1] with the following 
+information:
+
+* `fee_recipient`: an execution layer address where fees for the validator should go.
+* `gas_limit`: the value a validator prefers for the execution block gas limit.
+* `timestamp`: a recent timestamp later than any previously constructed `ValidatorRegistrationV1`
```
lightclient, line 58:
```suggestion
* `timestamp`: a recent timestamp later than any previously constructed `ValidatorRegistrationV1`.
```

specs/validator.md
lightclient: +1, since the VC needs to initiate the communication via `register_validator`; it makes sense that it tracks this information.

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+their building preferences are widely available in the external builder network.
+
+### Registration dissemination
+
+Beacon nodes are expected to keep track of the latest (determined by `timestamp`) registrations for each validator it 
+learns about and periodically submit them to upstream builder software using the 
+[`registerValidator`][register-validator-with-builder] endpoint of the standard 
+[APIs defined in the builder spec][builder-spec-apis].
+
+Registrations should be re-submitted frequently enough that any upstream builder software that comes online is aware of
```
lightclient, line 81: I think this is leaking some thoughts about a p2p mechanism. Right now builder software doesn't "come online" w.r.t. to the validator. The validator has to actively add the builder to their list of builders to use. So it's really only necessary submit the registration at that point or if other preferences in the registration are changed.

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+1. Call upstream builder software to get an `ExecutionPayloadHeader` with the
+   required data `slot`, `parent_hash` and `pubkey`, where:
+   * `slot` is the proposal's slot
+   * `parent_hash` is the value `state.latest_execution_payload_header.block_hash`
+   * `pubkey` is the propser's public key
+2. Assemble a `BlindedBeaconBlock` according to the process outlined in the [Bellatrix specs][bellatrix-specs] but with
+   the `ExecutionPayloadHeader` from the prior step in lieu of the full `ExecutionPayload`.
+3. The proposer signs the `BlindedBeaconBlock` and assembles a `SignedBlindedBeaconBlock` which is returned to the
+   upstream builder software.
+4. The upstream builder software in turn responds with the full `ExecutionPayload`. The proposer can use this payload
```
lightclient, line 111:
```suggestion
4. The upstream builder software responds with the full `ExecutionPayload`. The proposer can use this payload
```

specs/validator.md
```diff
@@ -0,0 +1,129 @@
...
+   to assemble a `SignedBeaconBlock` following the rest of the proposal process outlined in the
+   [Bellatrix specs][bellatrix-specs].
+
+#### Relation to local block building
+
+The external builder network offers a service for proposers that may from time to time fail to produce a timely block.
+Honest proposers who elect to use the external builder network **MUST** also build a block locally in the event the
```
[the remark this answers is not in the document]
lightclient: I think the timeout should be extremely short here as the builder should already have payload headers ready to go for the validator.

specs/validator.md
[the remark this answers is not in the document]
lightclient: image

lightclient (commented):
Overall looks really great; thank you @ralexstokes.

specs/validator.md
ralexstokes: think about all of the bytes we have wasted as a species on the "oxford comma"

specs/validator.md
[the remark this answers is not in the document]
g11tech: +1, once the signed blinded block is submitted and even if the rest api fails, for all intends and purposes this proposal is **over**. Any fallback will highly likely lead to slashing.

specs/validator.md
ralexstokes: my design goal here was to keep as much out of the VC as possible (bc it is the thing signing it should be even more secure imo than the beacon node and less code means fewer bugs) but you all make good arguments for having the VC do this; I'll update

specs/validator.md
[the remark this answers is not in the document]
ralexstokes: critical omission, will update

specs/validator.md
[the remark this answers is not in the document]
ralexstokes: I will add a note about a timeout with a configurable constant.

We can debate the actual value of the constant and it should also likely be informed by testnet operations.

There is also a question about how strictly a "honest" proposer needs to follow this restriction. It is best to have something concrete but there is a wider game a proposer can play to race the receipt of more valuable blocks against getting their particular block on chain.

specs/validator.md
ralexstokes: I'm talking about the high-level process here which is "end to end" and supposed to contrast w/ the default flow of getting an `ExecutionPayload` from some other execution engine. So it is correct to talk about the full payload and not the header.

specs/validator.md
ralexstokes: see previous comment. I'm talking end-to-end here

specs/validator.md
ralexstokes: I think it does though...

the validator client doesn't directly know about any builders, just the upstream beacon node.

we do leak some implementation details in that we have knowledge of a "relay" actor that sits b/t the beacon node and the network of builders (and is likely proxied by mux software).

and as long as we have this separation, there is the possibility that new builders will come online when a validator already is online and these new builders will need the validator registrations.

we can outsource this responsibility farther along the stack but then validators likely do not need to keep publishing their registrations and can just do so when there is a change (and we rely on stateful muxes or relays to track the data they have seen and also disseminate)

terencechain (approved):
lgtm!

specs/validator.md
```diff
@@ -0,0 +1,153 @@
...
+Honest proposers who elect to use the external builder network **MUST** also build a block locally in the event that the
+external builder network fails to provide a `SignedBuilderBid` in time in order to propagate the full
+`SignedBeaconBlock` during the proposer's slot.
+
+Honest proposers using the external builder network will give the builders a duration of
+`BUILDER_PROPOSAL_DELAY_TOLERANCE` before the external builder is considered to have hit the deadline and the external
+builder flow must be aborted in favor of a local build process.
+
+**WARNING**: Validators must be careful to not get slashed when orchestrating the duplicate building pathways.
+  See the [section on slashing][#proposer-slashing] for more information.
```
mkalinin, line 129:
```suggestion
  See the [section on slashing](#proposer-slashing) for more information.
```

specs/validator.md
```diff
@@ -0,0 +1,153 @@
...
+#### Relation to local block building
+
+The external builder network offers a service for proposers that may from time to time fail to produce a timely block.
+Honest proposers who elect to use the external builder network **MUST** also build a block locally in the event that the
+external builder network fails to provide a `SignedBuilderBid` in time in order to propagate the full
+`SignedBeaconBlock` during the proposer's slot.
+
+Honest proposers using the external builder network will give the builders a duration of
+`BUILDER_PROPOSAL_DELAY_TOLERANCE` before the external builder is considered to have hit the deadline and the external
+builder flow must be aborted in favor of a local build process.
```
mkalinin, line 126: What do you think about including fallback to a local build process to the steps outlined above? In this case the place from which the fallback becomes dangerous (due to slashing) could be clearly defined.

ralexstokes: I think it would obscure the flow outlined above but I'm open to it if many think it is clearer.

As far as I know most clients have some version of this implemented already and they have two distinct code paths so it maps directly to say "run these two tasks, race them against a deadline"

specs/validator.md
lightclient: Right, so my thinking has been that relays would be storing this information, because they need to validate the blocks their builders give them -- the only way to completely verify is to also store all the validator registrations. In that case, there would just need to be a mechanism for the relay to disseminate it to builders.

specs/validator.md
mkalinin: I am fine with current description as long as implementers clearly understand the last moment where they can safely sign off on a payload that is built locally.

specs/validator.md
```diff
@@ -0,0 +1,153 @@
...
+   * `parent_hash` is the value `state.latest_execution_payload_header.block_hash`
+   * `pubkey` is the propser's public key
+2. Assemble a `BlindedBeaconBlock` according to the process outlined in the [Bellatrix specs][bellatrix-specs] but with
+   the `ExecutionPayloadHeader` from the prior step in lieu of the full `ExecutionPayload`.
+3. The proposer signs the `BlindedBeaconBlock` and assembles a `SignedBlindedBeaconBlock` which is returned to the
+   upstream builder software.
+4. The upstream builder software responds with the full `ExecutionPayload`. The proposer can use this payload
+   to assemble a `SignedBeaconBlock` following the rest of the proposal process outlined in the
+   [Bellatrix specs][bellatrix-specs].
+
```
tbenr, line 116: I think we should mention somewhere here that the honest validator can start proposing using builder API **only after** the transition has been finalized. We agreed on this during the workshop.

specs/validator.md
tbenr: maybe it could be clearer if we refer `BUILDER_PROPOSAL_DELAY_TOLERANCE` specifically to bid providing.
something like 

... will give the builders a duration of `BUILDER_PROPOSAL_DELAY_TOLERANCE` to provide a `SignedBuilderBid` before the external builder is considered to have hit the deadline and the external

or also rename it to `BUILDER_BID_PROPOSAL_DELAY_TOLERANCE`?

this way it is clear the assumption that the fallback is no more an option once validator sign a blind block made with the bid.

specs/validator.md
ralexstokes: yeah I see where you are coming from, I'll update the text

ralexstokes:
ok I have addressed the feedback so far, unless there is more feedback soon I say we go ahead and merge it in

lightclient (approved):
LGTM

Issue #23, opened 2022-05-27
Clarify relationship b/t builder, relay and proposer


ralexstokes:
There is an ambiguity right now with the `pubkey` contained in the `SignedBuilderBid` and the `signature` included in that message.

We want to know which builder produced a given bid/block for accountability reasons and so the signed message should include an identifier (and ideally a signature) for the builder.

Proposers also want to be able to know (and importantly, prove to others) which relays they use when obtaining blocks from the external builder network. We ideally have a separate identifier *and* signature to bind a particular relay to a bid they forward to validators.

Right now, there is only one set of (pubkey, signature) in the `SignedBuilderBid` message and it is not clear in the specs which set of actors they refer to.

To remedy, one solution is:

1) Include a builder public key and signature in the `BuilderBid` message that signs over the `header` and `value` object.
2) Include a relay public key along with an outer signature in the `SignedBuilderBid` message.

And ideally the relay validates the signature from the builder and the proposer validates the signature from the relay *and* builder. Proposers could skip builder verification if they want, or perhaps do offline in situations where it becomes relevant.

Another solution that reduces the overhead here compared to the status quo is to make the builder opaque to the proposer (so that the `SignedBuilderBid` only has pubkey and signature for the relay) with an option to also have the relay sign over the builder identity (so include a second  pubkey for the builder in the `BuilderBid` message). If it is not clear from the messages exchanged in the builder APIs which builder was responsible for a block, then I think there should be a separate protocol put in place to increase accountability between builders and relays but if we go this route then this concern can be handled elsewhere.

metachris:
I feel like adding a builder signature would be good.

An alternative to the signature addition suggested above could be to embed both public keys into BuilderBid and have both signatures over the flat structure.

metachris:
I'd like to propose https://github.com/ethereum/builder-specs/pull/30/files as first step, adding both the builder and relay pk into the BuilderBid. This is useful for many cases, including not only relying on the relay pk pre-configured in mev-boost at startup.

PR #30, opened 2022-06-09
bid: change pubkey and include relay_pubkey

metachris:
Let's add both the builder and relay pk into the `BuilderBid`. This is useful for many cases, including not only relying on the relay pk pre-configured in mev-boost at startup.

### DIFF
--- a/types/bellatrix/bid.yaml
+++ b/types/bellatrix/bid.yaml
@@ -9,9 +9,12 @@ Bellatrix:
       value:
         $ref: "../../beacon-apis/types/primitive.yaml#/Uint256"
         description: "Payment in wei that will be paid to the `fee_recipient` account."
-      pubkey:
+      builder_pubkey:
         $ref: "../../beacon-apis/types/primitive.yaml#/Pubkey"
         description: "BLS public key of builder."
+      relay_pubkey:
+        $ref: "../../beacon-apis/types/primitive.yaml#/Pubkey"
+        description: "BLS public key of relay."
 
   SignedBuilderBid:
     type: object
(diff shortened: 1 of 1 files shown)


metachris:
wip, this is meant to facilitate discussions around https://github.com/ethereum/builder-specs/issues/23

metachris:
alright, let's close it for now and revisit if needed.

lightclient:
I think a source of confusion here is that the spec is overloading the builder actor in non-obvious ways. What I was hoping to achieve was the minimal interaction between a validator and someone with a block. In the 1:1 view where a validator connects directly to a builder, this terminology makes sense. However, in 1:N with mev-boost and relays it gets more confusing. 

My idea was that because the relay is the last point of contact for the validator, it absolves the complexity of the relay<>builder relationship. To the validator, a relay just looks like a builder. Therefore, it would be the relay's pubkey and signature in the "builder" bid.

I think we should think very hard about a change like this and ask why do we need both keys? It feels like a leak of the validator<>builder abstraction (using builder as a term from the validators perspective as noted above). It very well could be a hard requirement, but I would like to see a more convincing argument as to why we need the validator have the original builder's signature.

Also, forging the builder's signature is trivial if a relay and builder decide to collude. Right now, I don't think relay<>builder collusion affects the builder spec because it has black boxed that whole relationship. Once we add this additional field, I think that brings a whole host of new things into scope.

I agree with Alex that we should explore what's possible with the Relay API first. If the ultimate goal is to increase transparency and accountability, I think we can build a strong solution by just querying relays and/or builders. If we find that we need validators to weigh in in some way, we should probably revisit this idea.

PR #33, opened 2022-06-15
Clarify `fork_version` and `genesis_validator_root` for  domain

terencechain:
Further clarify what `fork_version` and `genesis_validator_root` should be for  `compute_domain(DOMAIN_APPLICATION_BUILDER)`. It was a little confusing at first until reading the spec code

### DIFF
--- a/specs/builder.md
+++ b/specs/builder.md
@@ -135,7 +135,8 @@ There are two types of data to sign over in the Builder API:
   and use the domain specified for beacon block proposals.
 * Builder API messages, e.g. validator registration, which should compute the
   signing root using [`compute_signing_root`][compute-root] with domain given by
-  `compute_domain(DOMAIN_APPLICATION_BUILDER)`.
+  `compute_domain(DOMAIN_APPLICATION_BUILDER)` with inputs 
+  `fork_version=GENESIS_FORK_VERSION` and `genesis_validators_root=Root()`.
 As `compute_signing_root` takes `SSZObject` as input, client software should
 convert in-protocol messages to their SSZ representation to compute the signing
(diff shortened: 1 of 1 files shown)


lightclient (approved):

ralexstokes (commented):
I would generally say to omit this kind of thing as, while it may not be clear from reading at first pass, does leverage the semantics of python we use liberally across the consensus specs.

I think the likelihood here is low but it does couple this part of this spec to any changes in the source -- if something about `compute_domain` changes then we will also have to change it here. So it seems to increase the maintenance burden to have the duplication.

That all being said if there isnt any other real opposition to this then I don't mine if it gets merged


edit: the content itself is fine, we could also inline it so that it is more like `compute_domain(DOMAIN_APPLICATION_BUILDER, fork_version=GENESIS_FORK_VERSION, genesis_validators_root=Root())`

terencechain:
> I would generally say to omit this kind of thing as, while it may not be clear from reading at first pass, does leverage the semantics of python we use liberally across the consensus specs.
> 
> I think the likelihood here is low but it does couple this part of this spec to any changes in the source -- if something about `compute_domain` changes then we will also have to change it here. So it seems to increase the maintenance burden to have the duplication.
> 
> That all being said if there isnt any other real opposition to this then I don't mine if it gets merged
> 
> edit: the content itself is fine, we could also inline it so that it is more like `compute_domain(DOMAIN_APPLICATION_BUILDER, fork_version=GENESIS_FORK_VERSION, genesis_validators_root=Root())`

No problem. I don't feel strongly about this. If there's no further opposition, I am happy to close this

PR #34, opened 2022-06-24
getHeader: 204 response if no header is available

metachris:


ralexstokes:
I think 204 fits for this use case

although I would also think it is fine to just return an error, rather than add more feature surface into the builder specs

wdyt?

metachris:
Also if the relay is not shipping blocks for a few epochs after the merge, I'd rather not return an error on every block (which will pollute the mev-boost and BN logs with errors as well). A clean no-content response would be much preferable imo.

lightclient (approved):
This makes a lot of sense to me.

PR #38, opened 2022-07-07
Merge transition delay

### LINKED ISSUE #32: Consider running builder software through the Merge testing
Remaining testnets before the Merge:

- ~Sepolia~
- Goerli

@metachris and I would like to consider CL clients running version `v0.1.0` of these specs during the upcoming Sepolia merge as a dress-rehearsal for later testnet merges and ultimately mainnet.

There is a trade-off here between shipping a safe merge with reduced code surface and deploying an untested software stack a few epochs into the finalized mainnet merge.

I'm raising this issue as a place for discussion of these various trade offs.

I'll open with the following extension to the `builder-specs`:

```
`MERGE_DELAY_EPOCHS` | `16` | `epoch`

Proposers **MUST** wait until `MERGE_DELAY_EPOCHS` epochs after the merge transition has been finalized before they can start querying the external builder network.

* NOTE: if the merge transition happens in epoch N and is finalized in epoch N+2, then proposers **MUST** not use the external builder network until epoch N + 2 + `MERGE_DELAY_EPOCHS`.
```

ralexstokes:
Fixes #32.

### DIFF
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -21,6 +21,7 @@
     - [Relation to local block building](#relation-to-local-block-building)
 - [How to avoid slashing](#how-to-avoid-slashing)
   - [Proposer slashing](#proposer-slashing)
+- [Responsibilites during the Merge transition](#responsibilites-during-the-merge-transition)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -46,6 +46,7 @@ and by extension the [Bellatrix consensus specs][bellatrix-specs].
 | - | - | - |
 | `EPOCHS_PER_VALIDATOR_REGISTRATION_SUBMISSION` | 1 | epoch(s)|
 | `BUILDER_PROPOSAL_DELAY_TOLERANCE` | 1 | second(s) |
+| `MERGE_DELAY` | 16 | epoch(s) |
 
 ## Validator registration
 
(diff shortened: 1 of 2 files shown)


mkalinin (approved):
sgtm!

metachris (approved):

ralexstokes:
> Will relays/mev-boost return 204 when querying for the execution payload header during the delay time?

this is the idea!

tbenr:
> > > Will relays/mev-boost return 204 when querying for the execution payload header during the delay time?
> > 
> > 
> > this is the idea!
> 
> Thanks! If CL handles the 204 properly and fallback to a local EL, is it not redundant to also add the additional logic from this PR?

It seems more an honest builder then a honest validator spec.

My concern is that CL discards pre finalized states. So CL can immediately know if we have a finalized execution payload but currently cant easily know when we finalized it the first time. Tracking it at runtime is easy but it wont survive a restart. So we need to write it on DB somewhere (or scan canonical blocks backward at startup). I mean, we can do that, just pointing out that it is not trivial to get that information ATM.

And if the main thing here is to have builders not serving blocks in that timeframe, seems a redundant behaviour.

ralexstokes:
@StefanBratanov @tbenr fair point on the semantics, the intent is to have some number we all agree upon

and the idea is that (1) block builders will not produce blocks during this time (2) relays will not serve blocks during this time (and will 204 if the header endpoint is called) and that (3) proposers won't bother calling the endpoint at this time

if one party just wants to rely on another party failing to do their job and "papering over" the error then that's fine but you are taking on the risk that they will in fact fail to do their job and then you could be exposing yourself to increased risk during the Merge transition -- you can decide how much risk to take on for an unprecedented software upgrade

this PR currently mixes the concerns across (1) (2) and (3) out of convenience as we don't have an "honest block builder" or "honest relay" spec yet although if you think it is worth while I can re-work this PR in that direction

ralexstokes:
@tbenr 

> My concern is that CL discards pre finalized states. So CL can immediately know if we have a finalized execution payload but currently cant easily know when we finalized it the first time. Tracking it at runtime is easy but it wont survive a restart. So we need to write it on DB somewhere (or scan canonical blocks backward at startup). I mean, we can do that, just pointing out that it is not trivial to get that information ATM.

I think we can discuss implementation on the next CL call and if there is enough appetite to go another route then we can change tactics here

StefanBratanov:
> if one party just wants to rely on another party failing to do their job and "papering over" the error then that's fine but you are taking on the risk that they will in fact fail to do their job and then you could be exposing yourself to increased risk during the Merge transition -- you can decide how much risk to take on for an unprecedented software upgrade

Agree actually. Ideally we would like 1, 2 and 3 to do their job properly to be fully robust. I guess my main concern was running this logic indefinitely. I suppose we could introduce a CLI flag which enables this logic and default it to false and it will be up to the users who run mev-boost/relay during merge transitions to enable it. Happy to discuss implementation details during the CL as you mentioned. :)

ullashb (approved):

djrtwo:
Right, @mkalinin. That conversation primarily left me with the impression that *just* finalizing the merge transition is likely the proper path here rather than the finality + delay. The subtle complexities that emerge with the delay are absolutely not worth it in my opinion as they just open the door for more errors and/or gameability. 

I would suggest just changing this to a finality trigger
and if we also want a delay, consider asking "honest" relays to add such a delay, but not 100% on that

ed738 (approved):

terencechain (approved):

ed738 (approved):

PR #39, opened 2022-07-19
clarify builder signing domain

tersec:


ralexstokes (commented):
@terencechain raised the same issue so if this is not clear then I'm happy to update

the idea is that this follows the semantics of the spec which inherits the python handling of "default" arguments

if we want to make it explicit then I'd suggest specifying the arguments inline like:

```python
compute_domain(DOMAIN_APPLICATION_BUILDER, fork_version=None, genesis_validators_root=None)
```

what do y'all think?

I generally don't want to stray too far from the pyspec style as I would like to retain the option to write more elaborate specs here in that style where we can, for example, generate conformance tests in a pretty  straightforward way

tersec:
This spec is not, however, written in Python pseudocode. Even after reading that's the intent, it's not something that I'd read into the actual prose.

Nothing about
> ### Signing
>
> All signature operations should follow the [standard BLS operations][bls]
> interface defined in `consensus-specs`.
>
> To assist in signing, we use a function from the [consensus specs][consensus-specs]: 
> [`compute_domain`][compute-domain]
>
> There are two types of data to sign over in the Builder API:
> * In-protocol messages, e.g. [`BlindedBeaconBlock`](#blindedbeaconblock), which
>   should compute the signing root using [`compute_signing_root`][compute-root]
>   and use the domain specified for beacon block proposals.
> * Builder API messages, e.g. validator registration, which should compute the
>   signing root using [`compute_signing_root`][compute-root] with domain given by
>   `compute_domain(DOMAIN_APPLICATION_BUILDER)`.
> As `compute_signing_root` takes `SSZObject` as input, client software should
> convert in-protocol messages to their SSZ representation to compute the signing
> root and Builder API messages to the SSZ representations defined
> [above](#containers).

suggests Python semantics.

If one wants to write something more explicit along the lines of
```python
compute_domain(DOMAIN_APPLICATION_BUILDER, fork_version=None, genesis_validators_root=None)
```
or even
```python
compute_domain(DOMAIN_APPLICATION_BUILDER)
```
which would directly invoke the exact/precise semantics of https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#compute_domain that's fine, but that's not what the current verbiage does.

tersec:
> Do you like the way I wrote it above or would you rather spell out the data like you did here?

I do like your suggested approach: https://github.com/ethereum/builder-specs/pull/39/commits/8f68a98d61fc1200d7ed93d4a9d58c7886e54054

ralexstokes (approved):
lgtm!

PR #41, opened 2022-07-21
Add Proposer config file specification

StefanBratanov:
Related to the proposer config file format discussion in https://github.com/flashbots/mev-boost/issues/154

### DIFF
--- a/schemas/mev_boost_proposer_config.json
+++ b/schemas/mev_boost_proposer_config.json
@@ -1,0 +1,66 @@
+{
+    "title": "MevBoostProposerConfig",
+    "type": "object",
+    "properties": {
+        "builder_relays_groups": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "string"
+            }
+        },
+        "proposer_config": {
+            "type": "object",
+            "patternProperties": {
+                "^0x[a-fA-F0-9]{96}$": {
+                    "type": "object",
+                    "properties": {
+                        "fee_recipient": {
+                            "$ref": "#/$defs/fee_recipient"
+                        },
+                        "builder": {
+                            "$ref": "#/$defs/builder"
+                        }
+                    }
+                }
+            },
+            "additionalProperties": false
+        },
+        "default_config": {
+            "type": "object",
+            "properties": {
+                "fee_recipient": {
+                    "$ref": "#/$defs/fee_recipient"
+                },
+                "builder": {
+                    "$ref": "#/$defs/builder"
+                }
+            }
+        }
+    },
+    "required": [
+        "default_config"
+    ],
+    "$defs": {
+        "fee_recipient": {
+            "type": "string",
+            "pattern": "^0x[a-fA-F0-9]{40}$"
+        },
+        "builder": {
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "type": "boolean"
+                },
+                "gas_limit": {
+                    "type": "string"
+                },
+                "relays": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            }
+        }
+    }
+}
--- a/schemas/validator_client_proposer_config.json
+++ b/schemas/validator_client_proposer_config.json
@@ -1,0 +1,72 @@
+{
+    "title": "ValidatorClientProposerConfig",
+    "type": "object",
+    "properties": {
+        "builder_relays_groups": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "string"
+            }
+        },
+        "proposer_config": {
+            "type": "object",
+            "patternProperties": {
+                "^0x[a-fA-F0-9]{96}$": {
+                    "type": "object",
+                    "properties": {
+                        "fee_recipient": {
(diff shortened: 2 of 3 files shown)


ralexstokes:
some questions/thoughts:

do we require all CLs to support this format? I don't know if it adds much to require them to do so and would require buy-in from all the CL devs

if we go down this road, then we will want to specify precedence for how data from this file interacts with the information in the `ValidatorRegistration`

it looks like the main addition here is a mapping from proposer key to set of acceptable relays -- should we consider just including this information in the `ValidatorRegistration`?

michaelsproul:
I'm loosely against making the proposer config format mandatory, but wouldn't be opposed to it being standardised here if there's sufficient uptake amongst clients.

In addition to being used to configure the relay assignments in `mev-boost` this config is also being used to set fee recipients and gas limits on validator clients. Presently there's no other standard way to set the gas limit on a validator client, although we could add a keymanager API as in this PR: https://github.com/ethereum/keymanager-APIs/pull/39

StefanBratanov:
> some questions/thoughts:
> 
> do we require all CLs to support this format? I don't know if it adds much to require them to do so and would require buy-in from all the CL devs
> 
> if we go down this road, then we will want to specify precedence for how data from this file interacts with the information in the `ValidatorRegistration`
> 
> it looks like the main addition here is a mapping from proposer key to set of acceptable relays -- should we consider just including this information in the `ValidatorRegistration`?

No, It is optional. I will try to reword it in a way to reflect that.

My understanding is that CL either communicates directly with a relay or mev-boost. Sending this additional information as part of the `ValidatorRegistration` will let other relays knows what relays a certain validator is using. I guess that is a problem?

StefanBratanov:
> I'm loosely against making the proposer config format mandatory, but wouldn't be opposed to it being standardised here if there's sufficient uptake amongst clients.
> 
> In addition to being used to configure the relay assignments in `mev-boost` this config is also being used to set fee recipients and gas limits on validator clients. Presently there's no other standard way to set the gas limit on a validator client, although we could add a keymanager API as in this PR: [ethereum/keymanager-APIs#39](https://github.com/ethereum/keymanager-APIs/pull/39)

Agree, it should be entirely optional. I think though standartization will help with setting up and maintaining builder architectures.

ralexstokes:
so I think what @metachris said here is worth calling out: https://github.com/flashbots/mev-boost/issues/154#issuecomment-1192340252

which implies we should remove all of the preference data from any standardized config and just focus on the unique concern here: mapping proposer public key to relay preferences

and personally at this point, it just feels like `mev-boost` specific configuration but if enough other people see value in standardizing this then I'm very open to it

StefanBratanov:
> so I think what @metachris said here is worth calling out: [flashbots/mev-boost#154 (comment)](https://github.com/flashbots/mev-boost/issues/154#issuecomment-1192340252)
> 
> which implies we should remove all of the preference data from any standardized config and just focus on the unique concern here: mapping proposer public key to relay preferences
> 
> and personally at this point, it just feels like `mev-boost` specific configuration but if enough other people see value in standardizing this then I'm very open to it

I agree. It would be worth spending more time on standardizing it only if people see the value in it. I would drop a suggestion to discuss this in the CL call agenda issue.

tersec:
1) What does yet another source (with what precedence versus other BN and VC sources of this information, that I believe all the CLs already have) of per-proposer fee recipient information provide? This isn't a particularly uniquely MEV-related concern.

2) Nimbus does not use a file like this for this, but to better support the dynamic key manager API, does so differently. This would not be a necessarily compelling per-recipient keymanager change, apropos (1), from a pure standalone perspective (for example, it doesn't scale with numbers of per-node validators well in modifying keys on nodes supporting thousands of validators either in reading or writing them: it requires additional caching on both ends to function well in that context). There are other critiques here, but fundamentally a MEV change is just ... an unexpected place to find this.

3) Until there are multiple public builder API relays for a network that's not mainnet (too high-stakes to test with initially), this is not feasible to test in the wild end-to-end, and at least as of a couple of weeks ago, the purely local relay test story was less than ideal.

metachris:
Is there any recent thoughts, conversations or developments about this?

beetrootkid:
It seems there will be at least 2 public relays for mainnet on the Merge. I also think, given some of the discussions around txn censorship and OFAC, some proposers will want their blocks to just come from the public mempool via the attached Execution Layer. I understand this can be achieved by simply leaving the relay url for a particular validator address empty and then the request reverts to the EL.

StefanBratanov:
> Is there any recent thoughts, conversations or developments about this?

I did have one more thought about this change and I think it makes more sense to add this information to the `mev-boost` docs rather than the builder specs explaining the config file from `mev-boost` point of view (only the relay configuration) and also mention that it is compatible with proposer configs of other clients (Teku, Prysm). What do you think?

ricardolyn:
as customers can dynamically change the relays on a staking platform, the MEV boost should be able to read from a remote URL, similar to what Teku does and automatically update the relayers for each validator. is there a plan to include that approach in this PR?

StefanBratanov:
closing this PR after discussion that let me to think that this is more `mev-boost` specification

PR #47, opened 2022-08-10
specify liveness failsafe for builder network

ralexstokes:
this PR specifies one version of a "circuit breaker" to reduce the scope of the failure domain in the event of liveness faults on chain

note to implementers: you can just count the distinct number of block roots from the `BeaconState` to get a stream of the proposed inputs ("M slots over a window of N slots")

### DIFF
--- a/specs/validator.md
+++ b/specs/validator.md
@@ -19,6 +19,11 @@
     - [Constructing the `BeaconBlockBody`](#constructing-the-beaconblockbody)
       - [ExecutionPayload](#executionpayload)
     - [Relation to local block building](#relation-to-local-block-building)
+  - [Liveness failsafe](#liveness-failsafe)
+    - [Failsafe initialization](#failsafe-initialization)
+    - [Failsafe activation](#failsafe-activation)
+    - [Failsafe recovery](#failsafe-recovery)
+    - [Selecting liveness parameters](#selecting-liveness-parameters)
 - [How to avoid slashing](#how-to-avoid-slashing)
   - [Proposer slashing](#proposer-slashing)
 - [Responsibilites during the Merge transition](#responsibilites-during-the-merge-transition)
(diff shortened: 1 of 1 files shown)


terencechain (approved):
Maybe it's worth mentioning that "missing slot" is not explicit only to the canonical chain, otherwise, a reorg’ing adversary could trivially disable mev-boost

terencechain:
> note to implementers: you can just count the distinct number of block roots from the `BeaconState` to get a stream of the proposed inputs ("M slots over a window of N slots")

This will count orphaned blocks as missing, no?

ralexstokes:
I feel like the outcome on the call was that we would just stick to the canonical chain as it would be simpler to implement -- just look at the canonical beacon state

and to mitigate any sort of reorg'ing attacks we now want to widen the window to trigger the breaker

and if the current suggested range for `ALLOWED_FAULTS` feels too small we can bump it up (and possibly also widen the rolling window)

terencechain:
> I feel like the outcome on the call was that we would just stick to the canonical chain as it would be simpler to implement -- just look at the canonical beacon state


I don't have an issue with this. For prysm it's equally simple looking from the canonical chain's perspective or combining canonical and forked chain's perspectives. We'll stick with counting missing slots as they are truly missed. I think client diversity is a nice to have here

ralexstokes:
yes, I think the first goal is: "clients have *something* implemented" if only to dissuade potential actors who would abuse this attack vector

the second goal is: "do clients in aggregate implement something that is hard to attack?" -- this is where it matters a bit who is doing what bc if someone has a more sensitive trigger then it could selectively be used to take the builder network offline

specs/validator.md
```diff
@@ -129,6 +134,46 @@ the deadline and the external builder flow must be aborted in favor of a local b
...
+If the circuit breaker is active and the total number of missing slots during the inspection window becomes less than
+or equal to `ALLOWED_FAULTS`, the circuit breaker deactivates. The beacon client continues to monitor the chain on a
+rolling basis to consider re-activation.
+
+#### Selecting liveness parameters
+
+Beacon clients select randomized values from the following ranges when initializing the circuit breaker (so at boot time
+and once for each unique boot).
+
+* `ALLOWED_FAULTS`: between `1` and `SLOTS_PER_EPOCH // 2`
```
mkalinin, line 174: Wouldn't `1` be too low?

In combination with `2 * SLOT_PER_EPOCH` it will activate the circuit breaker after two epochs with missed slots. Although, it *seems* unlikely on mainnet but not unrealistic

specs/validator.md
```diff
@@ -129,6 +134,46 @@ the deadline and the external builder flow must be aborted in favor of a local b
...
+
+#### Failsafe initialization
+
+When a beacon client first boots, the circuit breaker is instantiated with the liveness parameters and is **not**
+active.
+
+#### Failsafe activation
+
+The circuit breaker becomes active if a liveness fault is observed locally which is determined by considering the beacon
+chain over a rolling window of `FAULT_INSPECTION_WINDOW` slots.
```
rolfyone, line 158: so basically once the node is 'in sync', it will start counting missed slots, and missed slots exceeding `ALLOWED_FAULTS` within a defined window deem that we should stop using the builder network...

Rather than random, it may be more appropriate to have actual configured values to use, that way it could be tweaked so that builder can work on 'test' networks that have more missed blocks...

eg mainnet it might be unusual to see 2 missed blocks in 64 slots, but that's basically business as usual on several testnets...

Doesn't this also suggest that the builder can be taken out through no fault of it's own? If a couple of validators were offline at an inopportune moment, then the builder network could be completely shut down for the period defined by the inspection network?

Edge case would be allowed_faults = 1, fault_inspection_window = 64, then 2 missed slots in 2 epochs would stop you using builders until the first one clears... Then another happens, and you're still locked out, which could result in builder not being used for a prolonged period? So 1 missed block in 32 slots is enough in this case to ensure builder is never used?

specs/validator.md
```diff
@@ -129,6 +134,46 @@ the deadline and the external builder flow must be aborted in favor of a local b
...
+any building happens locally).
+
+The specifics of instantiating, activating and deactivating the circuit breaker are given below. There are references
+to two parameters `ALLOWED_FAULTS` and `FAULT_INSPECTION_WINDOW` and the selection of concrete values for these
+constants is discussed after specifying the mechanics.
+
+#### Failsafe initialization
+
+When a beacon client first boots, the circuit breaker is instantiated with the liveness parameters and is **not**
+active.
```
g11tech, line 153: if I am reading it correctly then: circuit breaker is recommended to be "NOT" active on a start/restart, if thats the case then i feel it should be opposite because its highly likely that in an unhealthy network the nodes will be rebooted and should follow the safest/most conservative route. 
This could be client specific but still the recommendation should be reverse.

tersec:
In terms of reducing the scope of the failure domain in the event of liveness faults on chain, given that https://github.com/remyroy/ethstaker/blob/main/MEV-relay-list.md documents multiple relay networks, it's not obvious that it's more effectively centralized than a still-near-Geth-monoculture on the EL side.

The implied calculation is that the most likely inference from a bunch of missed slots is that (a) a large portion of the network is using the builder API; and (b) the builder API is less reliable in a correlated-across-CL-nodes way than the engine API. (a) hasn't yet proven as true as some earlier predictions, and I'm not sure (b) is useful to bet on -- even if one relay network has problems, it's reasonable to try another, and `mev-boost` can already be configured to try multiple relay networks.

While local EL infrastructure for the engine API is definitely less centralized per se, that doesn't protect against already-witnessed situations where all or most EL instances of a certain type across a network fail to propose good blocks in some situation.

Rather, the builder and engine APIs act a kind of diversity in themselves, and given that the former isn't simply a FB interface anymore, responding to network trouble by decreasing decentralization seems risky.

ralexstokes:
closing in lieu of #95

PR #49, opened 2022-08-15
Deprecate Builder specs in lights of Flashbots adopting OFAC sanctions

Giulio2002:
## Rationale

Flashbots has been found to censor transactions in their bundles, and It is just a matter of time for when they are going to do the same in MEV-Boost. I believe, it is an attack on Ethereum and Decentralization as a whole and needs to be tackled somehow. this is just the first shot.

## Additions

This PR adds a couple of deprecation notes on the builder specs and momentarily deprecates it. flashbots can use their own closed source code to do this but I believe that the Ethereum community should not let users use MEV-Boost. **Note: this deprecation is meant to be momentary, and will by no mean extend for too much time.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ In this repository:
 * [Honest validator specification][validator-spec]
 
 
-### Builder API is currently deprecated
+### ⚠️⚠️⚠️ Builder API is currently deprecated ⚠️⚠️⚠️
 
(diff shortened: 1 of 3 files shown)


lightclient:
If you actually want to have this PR merged, I recommend making your case in the PR description. There is also discussions happening in `#block-construction` on the Eth R&D discord on this topic that you may want to participate in.

PR #58, opened 2022-12-01
Builder spec changes for Capella & EIP-4844

jimmygchen:
## Description

- Updated `ExecutionPayload` & `BeaconBlockBody`  types to align with EIP-4844 specs. (Depends on types from this Beacon API PR: https://github.com/ethereum/beacon-APIs/pull/271)
- Add `blob_kzg_commitments` to `getExecutionPayloadHeader` response (`SignedBuilderBid`), so that validator can include it in the block and create a signature over the blinded block.
- Update `submitBlindedBlock` endpoint for 4844, that includes `blobs_sidecar` in the response (`ExecutionPayloadAndBlobsSidecar`).

## Changes for EIP-4844

```mermaid
sequenceDiagram
    participant validator
    participant consensus
    participant mev_boost
    Title: Block Proposal with Builder API (EIP-4844)
    validator->>consensus: GET v2/validator/blinded_blocks/{slot}
    consensus->>mev_boost: GET v1/builder/header/{slot}/{parent_hash}/{pubkey}
    mev_boost-->>consensus: GET v1/builder/header/{slot}/{parent_hash}/{pubkey} response
    Note left of mev_boost: add `blob_kzg_commitments` to response (`SignedBuilderBid`)
    consensus-->>validator: GET v2/validator/blinded_blocks/{slot} response
    Note left of consensus: add `blob_kzg_commitments` to response
    Note over validator: include `blob_kzg_commitments` in block and sign the header
    validator->>consensus: POST beacon/blinded_blocks
    consensus->>mev_boost: POST /eth/v2/builder/blinded_blocks
    mev_boost-->>consensus: POST /eth/v2/builder/blinded_blocks response
    Note left of mev_boost: revealed payload will have to include `blobs_sidecar` and kzg proof
    Note over consensus: construct block and broadcast block
```


apis/builder/blinded_blocks.yaml
```diff
@@ -19,9 +19,15 @@ post:
...
             - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
+            - $ref: "../../builder-oapi.yaml#/components/schemas/Capella.SignedBlindedBeaconBlock"
+            - $ref: "../../builder-oapi.yaml#/components/schemas/Eip4844.SignedBlindedBeaconBlock"
         examples:
           bellatrix:
             $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBlindedBeaconBlock"
+          capella:
+            $ref: "../../builder-oapi.yaml#/components/examples/Capella.SignedBlindedBeaconBlock"
+          eip4844:
+            $ref: "../../builder-oapi.yaml#/components/examples/Eip4844.SignedBlindedBeaconBlock"
```
jimmygchen, line 30: Not sure if we should support eip-4844 version  for the v1 API if we plan to use `submitBlindedBlockV2`?

0xGabi: That's a good point. Looking the [versioning section of the `execution-apis`](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#versioning) my understanding is that we should treat each version as independent. I vote for removing 4844 on v1.

types/eip4844/block_and_blobs_sidecar.yaml
```diff
@@ -0,0 +1,39 @@
+Eip4844:
```
0xGabi, line 1: I'm not convinced of having this types definitions on the builder spec. I think we should move them to the [`beacon-apis`](https://github.com/ethereum/beacon-APIs) repo. I see this types definitions being useful to more dependencies of the `beacon-apis`.

apis/builder/blinded_blocks.yaml
0xGabi: What do you think of including Capella as part of the `submitBlindedBlockV2`? And remove Capella form v1 as well.

0xGabi (changes requested):
Great work on adding the `blobs_sidecar`, I believe we are getting close to mark the PR ready for review.

I had some general thoughts about writing the spec:

- While looking through the [`execution-apis` spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md) I start wondering if we want to further clarify the changes between `submitBlindedBlockV1` and `submitBlindedBlockV2`.

- If we follow the same [versioning criteria](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#versioning) of the execution API. We may need to version the structures (i.e. `SignedBuilderBidV2`) and as a consequence have a `getHeaderV2`. I'm not sure we want to follow that path but wanted to bring it for discussion.

jimmygchen:
> What do you think of including Capella as part of the submitBlindedBlockV2? And remove Capella form v1 as well.
I'm not sure if we want to include submitBlindedBlockV2 in V2 since it's not strictly required for Capella, and I think we'd need to be able to ship withdrawals / Capella if 4844 doesn't get implemented in time.

types/eip4844/block_and_blobs_sidecar.yaml
jimmygchen: There's currently no dependencies in Beacon API, so adding it there means we'd end up having unused schema in the repo and could get accidentally deleted. I'm all for moving it there if it's required in the Beacon API.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "beacon-apis"]
 	path = beacon-apis
-	url = git@github.com:ethereum/beacon-APIs.git
+	url = git@github.com:jimmygchen/beacon-APIs.git
```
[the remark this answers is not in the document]
jimmygchen: Done

apis/builder/blinded_blocks.yaml
jimmygchen: > That's a good point. Looking the [versioning section of the `execution-apis`](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#versioning) my understanding is that we should treat each version as independent. I vote for removing 4844 on v1.

I kept the 4844 here because we haven't really discussed about deprecating this, and making it a requirement for clients to use V2. I think it would be great to get some feedback from client teams on this.

> What do you think of including Capella as part of the submitBlindedBlockV2? And remove Capella form v1 as well

I'm not sure if we want to include submitBlindedBlockV2 in V2 since it's not strictly required for Capella, and I think we'd need to be able to ship withdrawals / Capella if 4844 doesn't get implemented in time.

jimmygchen:
> * While looking through the [`execution-apis` spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md) I start wondering if we want to further clarify the changes between `submitBlindedBlockV1` and `submitBlindedBlockV2`.
> * If we follow the same [versioning criteria](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#versioning) of the execution API. We may need to version the structures (i.e. `SignedBuilderBidV2`) and as a consequence have a `getHeaderV2`. I'm not sure we want to follow that path but wanted to bring it for discussion.

Great point, I'm not sure tbh, and I'd like to hear some more discussions on this too. There were some discussions on discord earlier, and it looks like we don't need a new version just for new fields, however in this case we're changing the endpoint to return a  `SignedBeaconBlockAndSidecar` instead of `ExecutionPayload`, so it's a breaking change.
https://discord.com/channels/595666850260713488/1031999860997619843/1038023245900820531

apis/builder/header.yaml
```diff
@@ -50,14 +50,20 @@ get:
...
                  - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBuilderBid"
+                 - $ref: "../../builder-oapi.yaml#/components/schemas/Capella.SignedBuilderBid"
+                 - $ref: "../../builder-oapi.yaml#/components/schemas/Eip4844.SignedBuilderBid"
           examples:
             bellatrix:
               $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBuilderBid"
+            capella:
+              $ref: "../../builder-oapi.yaml#/components/examples/Capella.SignedBuilderBid"
+            eip4844:
+              $ref: "../../builder-oapi.yaml#/components/examples/Eip4844.SignedBuilderBid"
```
jimmygchen, line 66: I'm having some issues with lint here complaining about these examples matching multiple schemas, and not sure what's the best way to resolve this.

I think the json examples are matching multiple schemas because the schema allows for additional properties. I think we could set `additionalProperties` to `false` so that it doesn't match all, but I haven't seen this approach used anywhere else in the API repos.

```
~/builder-specs/examples/bellatrix/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data

~/builder-specs/examples/capella/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data

~/builder-specs/examples/eip4844/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data
```

apis/builder/blinded_blocks_v2.yaml
```diff
@@ -0,0 +1,58 @@
+post:
```
metachris, line 1: from the discussions i've had, there seems to be a preference for reusing the existing API call, rather than introducing a new V2 one.

jimmygchen: Hey thanks for the insight @metachris. I've seen some discussion on discord earlier too. My understanding was that the version header is used for determining the payload structure, so clients should be able to parse the same payload type from different forks. It make sense to me to keep the same API if the response type is the same, i.e. if we return `Eip4844.ExeuctionPayload`.

However I think we'd also want to include blobs and the kzg proof, so it would be more convenient to change the response type to `SignedBeaconBlockAndBlobsSidecar` (with the revealed `ExecutionPayload` wrapped within the block), and the clients can just broadcast the block after receiving it, rather than having to retrieve the sidecar and construct the beacon block. Does this make sense? I haven't seen an API allowing a completely different response type, so I thought we may want to increment the version, but keen to hear your thoughts! 

(cc: @ajsutton)

metachris:
Just wanted to say thanks for the contribution! 🚀  
Would love to understand more about where this is coming from.

pinging @lightclient @jtraglia @lightclient @terencechain @tbenr @ajsutton (and others) for review and feedback

apis/builder/header.yaml
jimmygchen: I don't see any easy way around without adding `required` for new properties everywhere, which is a bit ugly, so I went with using `anyOf` instead - if anyone has a better solution for this please let me know!

tbenr:
Regarding submitBlindedBlock, It makes sense to me to have a V2 returning the entire`SignedBeaconBlockAndBlobsSidecar`. It reduces complexity and the overhead is minimal since `ExecutionPayload` and `BlobsSidecar` are the biggest objects we need to get anyway.

types/capella/bid.yaml
```diff
@@ -0,0 +1,23 @@
...
+    properties:
+      header:
+        $ref: "../../beacon-apis/types/capella/execution_payload.yaml#/Capella/ExecutionPayloadHeader"
+        description: "`ExecutionPayloadHeader` to use in block proposal."
+      value:
+        $ref: "../../beacon-apis/types/primitive.yaml#/Uint256"
+        description: "Payment in wei that will be paid to the `fee_recipient` account."
+      pubkey:
+        $ref: "../../beacon-apis/types/primitive.yaml#/Pubkey"
+        description: "BLS public key of builder."
```
ralexstokes, line 14: I think we can define a `BuilderBidCommon` schema to reduce the repetition here

examples/capella/execution_payload.json
```diff
@@ -0,0 +1,26 @@
...
+      "timestamp": "1",
+      "extra_data": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "base_fee_per_gas": "1",
+      "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "transactions": [
+        "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
+      ],
+      "withdrawals": [
+        "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
+      ]
```
ralexstokes, line 23: ideally we have an actual hex-encoded withdrawal here

examples/eip4844/execution_payload.json
```diff
@@ -0,0 +1,27 @@
...
+      "timestamp": "1",
+      "extra_data": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "base_fee_per_gas": "1",
+      "excess_data_gas": "1",
+      "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "transactions": [
+        "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
+      ],
+      "withdrawals": [
+        "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
```
ralexstokes, line 23: same on withdrawal example here

types/eip4844/block_and_blobs_sidecar.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: string
+    format: hex
+    pattern: "^0x[a-fA-F0-9]{96}$"
+    description: "An aggregated KZG proof. Same check as `KZGCommitment`"
+
+  Blob:
+    type: string
+    format: hex
+    pattern: "^0x[a-fA-F0-9]{262144}$"
+    description: "A blob is `FIELD_ELEMENTS_PER_BLOB * size_of(BLSFieldElement) = 4096 * 32 = 131072` bytes (`DATA`) representing a SSZ-encoded Blob as defined in EIP-4844"
```
ralexstokes, line 38: what do you think about moving `SignedBeaconBlockAndBlobsSidecar` to the `beacon-apis` and then just referencing them here?

apis/builder/header.yaml
```diff
@@ -50,14 +50,20 @@ get:
             properties:
               version:
                 type: string
-                enum: [ bellatrix ]
+                enum: [ bellatrix, capella, eip4844 ]
                 example: "bellatrix"
               data:
-                oneOf:
+                anyOf:
```
ralexstokes, line 56: not sure on the semantics of `oneOf` vs `anyOf` but it sounds like we want `oneOf` here

apis/builder/blinded_blocks_v2.yaml
ralexstokes: > I haven't seen an API allowing a completely different response type

this should be fine -- we should be able to just add the right type to the v1 endpoint

let's drop v2 here and extend v1 to also support a new type that contains the `ExecutionPayload` and the `BlobsSidecar`, calling it something like `ExecutionPayloadWithBlobsSidecar`

ralexstokes (changes requested):
excellent work!

I left some questions and feedback -- this is a great start :)

This PR is fine as is but something to keep in mind for the future is that it is much easier to review smaller PRs, so this one could easily be a PR just for capella changes and then a separate one for 4844 changes

apis/builder/header.yaml
jimmygchen: Yes, it should be `oneOf`, I had a bit of issues getting this to work earlier (noted in [this comment](https://github.com/ethereum/builder-specs/pull/58/files/ea928bf1b8db6221bf729b935a889b890cff279c#r1037860896)) - perhaps we'll need to make all new properties in schema versions `required` in order for the examples pass validation::

> I'm having some issues with lint here complaining about these examples matching multiple schemas, and not sure what's the best way to resolve this.
> I think the json examples are matching multiple schemas because the schema allows for additional properties. I think we could set `additionalProperties` to `false` so that it doesn't match all, but I haven't seen this approach used anywhere else in the API repos.

```
~/builder-specs/examples/bellatrix/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data

~/builder-specs/examples/capella/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data

~/builder-specs/examples/eip4844/signed_builder_bid.json
 4:13  error  oas3-valid-media-example  "data" property must match exactly one schema in oneOf  value.data
```

> I don't see any easy way around without adding required for new properties everywhere, which is a bit ugly, so I went with using anyOf instead - if anyone has a better solution for this please let me know!

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "beacon-apis"]
 	path = beacon-apis
-	url = git@github.com:ethereum/beacon-APIs.git
+	# update to ethereum repo once PR is merged: https://github.com/ethereum/beacon-APIs/pull/271
+	url = git@github.com:jimmygchen/beacon-APIs.git
```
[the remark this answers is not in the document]
jimmygchen: done

types/eip4844/block_and_blobs_sidecar.yaml
jimmygchen: @0xGabi made a comment about this [here](https://github.com/ethereum/builder-specs/pull/58#discussion_r1037828181) as well - I wasn't sure about this as it's not used in `beacon-apis` (and may lead to validation warnings if there's one), but happy to move this since we've got more arguments for moving over to `beacon-apis`

apis/builder/blinded_blocks.yaml
jimmygchen: we can close this now, given we're not going ahead with V2

apis/builder/header.yaml
jimmygchen: Changed to `oneOf`. Got this working by adding `additioanlProperties: false` to `ExecutionPayloadHeader` and `BuilderBid` types.

types/eip4844/block_and_blobs_sidecar.yaml
jimmygchen: Done.

types/eip4844/block_and_blobs_sidecar.yaml
jimmygchen: moved to beacon APIs

apis/builder/blinded_blocks_v2.yaml
jimmygchen: Done, thanks!

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "beacon-apis"]
 	path = beacon-apis
-	url = git@github.com:jimmygchen/beacon-APIs.git
+	# update to ethereum repo once PR is merged: https://github.com/ethereum/beacon-APIs/pull/271
```
0xGabi, line 3:
```suggestion
```

0xGabi (approved):
Well done. I like `ExecutionPayloadAndBlobSidecar`.

types/capella/bid.yaml
jimmygchen: Done, thanks

examples/capella/execution_payload.json
jimmygchen: ah nice catch, updated now - just realised `withdrawals` is a list of objects.. wonder why validation didn't pick up 🤔

examples/eip4844/execution_payload.json
jimmygchen: Done

jimmygchen:
Thanks for the review @ralexstokes @0xGabi! 🙏  I've addressed your comments, please review again.

examples/capella/execution_payload.json
```diff
@@ -19,8 +19,13 @@
         "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
       ],
       "withdrawals": [
-        "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
+        {
```
0xGabi, line 22: 👍

0xGabi (approved):
LGTM

metachris:
I think it would make sense to split this PR into two -- one for Capella changes, and another one for EIP4844 changes? (The 4844 PR could point to the Capella PR to reuse anything that's already in there)

examples/eip4844/signed_beacon_block_and_blobs_sidecar.json
```diff
@@ -0,0 +1,176 @@
...
+            "timestamp": "1",
+            "extra_data": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+            "base_fee_per_gas": "1",
+            "excess_data_gas": "1",
+            "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+            "transactions": [
+              "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
+            ],
+            "withdrawals": [
+              "0x02f878831469668303f51d843b9ac9f9843b9aca0082520894c93269b73096998db66be0441e836d873535cb9c8894a19041886f000080c001a031cc29234036afbf9a1fb9476b463367cb1f957ac0b919b69bbc798436e604aaa018c4e9c3914eb27aadd0b91e10b18655739fcf8c1fc398763a9f1beecb8ddc86"
```
g11tech, line 147: withdrawals json here looks incorrect

jimmygchen: Thanks @g11tech that's fixed now!

jimmygchen:
> I think it would make sense to split this PR into two -- one for Capella changes, and another one for EIP4844 changes? (The 4844 PR could point to the Capella PR to reuse anything that's already in there)

Hi @metachris @ralexstokes I've just done that now, please see the two separated PRs
- Capella #60 
- EIP-4844 #61 

Closing this PR.

PR #60, opened 2023-01-03
Add and update types for Capella

jimmygchen:
Builder spec changes for Capella, separated from PR #58.

## Description

- Updated `ExecutionPayload` & `BeaconBlockBody`  types to align with Capella.


jimmygchen:
@ralexstokes @0xGabi would you mind reviewing this please? I've addressed all comments from the previous [PR](https://github.com/ethereum/builder-specs/pull/58). This PR contains changes for Capella only.

g11tech (approved):
lgtm!

apis/builder/header.yaml
```diff
@@ -45,19 +45,21 @@ get:
...
-            type: object
-            properties:
-              version:
-                type: string
-                enum: [ bellatrix ]
-                example: "bellatrix"
-              data:
-                oneOf:
-                 - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBuilderBid"
+            # discrimantor is required for schema validation to pass, as the examples could match more than one schema.
```
jimmygchen, line 48: @ralexstokes thanks for triggering the build - I was having issues earlier with schema validation as well after splitting the PR, as it revealed some problems with schema validation that was shadowed earlier.

I've added [one more commit](https://github.com/ethereum/builder-specs/pull/60/commits/702791c76f44bcbfa30d2d551d1bd999cdb66de6) to fix this and the lint is now passing. The schema validation was failing because as all fields are optional by default, so the examples match all schemas, hence `oneOf` wasn't satisfied. 

So I've added the use of `disciminator` to help it detect the schema, and I had to add a wrapper type GerHeaderResponse, as the discriminator must be used at the same level as `oneOf`.

specs/builder.md
```diff
@@ -121,6 +123,25 @@ class SignedBlindedBeaconBlock(Container):
...
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data
+    graffiti: Bytes32
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    sync_aggregate: SyncAggregate
+    execution_payload_header: ExecutionPayloadHeader
```
rkapka, line 141: Don't we need `ExecutionPayloadHeaderCapella, `SignedBuilderBidCapella` and `BuilderBidCapella`?

types/capella/bid.yaml
```diff
@@ -0,0 +1,30 @@
+Capella:
+  BuilderBid:
+    description: "The `BuilderBid` object from the Builder API Capella spec."
+    allOf:
+      - $ref: '../bellatrix/bid.yaml#/Bellatrix/BuilderBidCommon'
+      - type: object
+        properties:
+          header:
+            $ref: "../../beacon-apis/types/capella/execution_payload.yaml#/Capella/ExecutionPayloadHeader"
```
rkapka, line 9: This PR doesn't contain the file `/types/capella/execution_payload.yaml`

specs/builder.md
```diff
@@ -121,6 +123,25 @@ class SignedBlindedBeaconBlock(Container):
     signature: BLSSignature
 ```
 
+#### Capella
+
+##### `BlindedBeaconBlockBody`
+
+```python
+class BlindedBeaconBlockBody(Container):
```
rkapka, line 131: Can we have the same name as for Bellatrix?

specs/builder.md
rkapka, line 141: Don't we need `ExecutionPayloadHeaderCapella`, `SignedBuilderBidCapella` and `BuilderBidCapella`?

types/capella/bid.yaml
jimmygchen: Hi @rkapka the beacon-apis directory is a git [submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) and references the [`Beacon-APIs` repo](https://github.com/ethereum/beacon-APIs).

The submodule version has also been updated as part of this PR, which now includes the capella types. and can be found [here](https://github.com/ethereum/beacon-APIs/tree/77907ea6db2a50bbf6efd44a6b15ea19c2a3b521)

jimmygchen: you can run `git submodule update --init` to clone/fetch the submodules.

specs/builder.md
jimmygchen: I think so - this is the same other ethereum specs (e.g. [consensus-specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#containers), the type names remain the same across all forks, by convention.

specs/builder.md
jimmygchen: Good catch, I'll add them in - thanks!

jimmygchen: Just comparing it against the consensus specs again, I think we probably want to avoid duplicating the consensus spec types here (since the top of this page already mentions that these types are supplementary to the definitions in `[consensus-specs](https://github.com/ethereum/consensus-specs)`, duplicating here adds maintenance effort and we should use the consensus spec as single source of truth.

I'll add the missing BuilderBid type for Capella and add links for the consensus spec types.

jimmygchen: ^^ sorry please ignore 😂 , they are not in `consensus-specs`, I'll add them in

rkapka: Thanks for the update! You added several `# [Modified in EIP-4844]` comments. I think they should read `# [Modified in Capella]`.

jimmygchen: thanks! been context switching between the two too much 😅

rkapka:
We should merge this soon :-)

specs/builder.md
ralexstokes: the names shall match _but_ we should consider splitting the `bellatrix` material into a `specs/bellatrix.md` and the `capella` material into a `specs/capella.md` file

similar to how we do this w/ the engine API: https://github.com/ethereum/execution-apis/tree/main/src/engine

ralexstokes: I can follow up and do this myself or you can do it before we merge this PR, I'm ok either way

specs/builder.md
```diff
@@ -18,6 +18,8 @@
       - [`BlindedBeaconBlockBody`](#blindedbeaconblockbody)
       - [`BlindedBeaconBlock`](#blindedbeaconblock)
       - [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
+    - [Capella](#capella)
+      - [`BlindedBeaconBlockBody`](#blindedbeaconblockbody-1)
```
ralexstokes, line 22: looks like this is incomplete for some reason

examples/capella/signed_blinded_beacon_block.json
```diff
@@ -0,0 +1,158 @@
...
+      "parent_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "state_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+      "body": {
+        "randao_reveal": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505",
+        "eth1_data": {
+          "deposit_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+          "deposit_count": "1",
+          "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
+        },
+        "graffiti": "0xdeadbeefc0ffee",
```
ralexstokes, line 15: this field is a `Bytes32` (so fixed size of 32 bytes)

examples/capella/signed_blinded_beacon_block.json
```diff
@@ -0,0 +1,158 @@
...
+                "epoch": "1",
+                "root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
+              }
+            }
+          }
+        ],
+        "deposits": [
+          {
+            "proof": [
+              "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
```
ralexstokes, line 104: there may be a realistic example from the `beacon-apis` -- a deposit proof I believe is 6 (or 5) leaves

types/bellatrix/bid.yaml
```diff
@@ -1,18 +1,24 @@
...
 
+  BuilderBid:
+    description: "The `BuilderBid` object from the Builder API Bellatrix spec."
+    allOf:
+      - $ref: '#/Bellatrix/BuilderBidCommon'
+      - type: object
+        properties:
+          header:
+            $ref: "../../beacon-apis/types/bellatrix/execution_payload.yaml#/Bellatrix/ExecutionPayloadHeader"
+            description: "`ExecutionPayloadHeader` to use in block proposal."
```
ralexstokes, line 20:
```suggestion
      - type: object
        properties:
          header:
            $ref: "../../beacon-apis/types/bellatrix/execution_payload.yaml#/Bellatrix/ExecutionPayloadHeader"
            description: "`ExecutionPayloadHeader` to use in block proposal."
      - $ref: '#/Bellatrix/BuilderBidCommon'
```

I think this is the backwards-compatible thing

types/capella/bid.yaml
ralexstokes, line 9: same thing on field ordering here

apis/builder/header.yaml
ralexstokes: it looks like we are :)

this PR changed the linter:
https://github.com/ethereum/beacon-APIs/pull/223


I can update `main` to use the new linter

specs/builder.md
jimmygchen: @ralexstokes I'm happy to do this. I'm wondering if we want one folder per fork as there are two files `builder.md` and `validator.md`?  e.g. `specs/bellatrix/` and `specs/capella`, similar to [`consensus-specs`](https://github.com/ethereum/consensus-specs/tree/dev/specs). There aren't any changes to `validator.md` in capella though, so splitting it too granular may make it difficult to find information.

specs/builder.md
jimmygchen: Thanks - forgot to regenerate TOC earlier, I've pushed the fix

examples/capella/signed_blinded_beacon_block.json
jimmygchen: Thanks, fixed both capella and bellatrix examples

examples/capella/signed_blinded_beacon_block.json
jimmygchen: Updated example

types/bellatrix/bid.yaml
jimmygchen: Committed your suggestion, makes sense to me I think. For my understanding, so this just makes sure the "common" schema doesn't get overriden by the other right?

apis/builder/header.yaml
jimmygchen: @ralexstokes I've created a PR here to update the linter to redocly
https://github.com/ethereum/builder-specs/pull/64

ralexstokes: excellent! in this case I'd just follow the structure of the beacon apis where I think you can drop this `discriminator` thing (like you had before)

specs/builder.md
ralexstokes: yeah the folder structure works! I want to refactor some of the validations into more of a CL pyspec style so we may get more files soon(tm)

I'd rather not clutter this one file and lay the foundation for the striation by fork, even if we end up with only one additional file in the new folder

types/bellatrix/bid.yaml
ralexstokes: I was just trying to indicate the relative ordering of the accepted object

JSON objects don't have order within one "key scope" so its minor but I think what you had would put the header _after_ all of the "common" stuff and I believe the header was given first

I intend for my suggestion to put them in the same order

ralexstokes (commented):
nice updates! I left some more comments but this is very close to merge

apis/builder/blinded_blocks.yaml
```diff
@@ -19,9 +19,12 @@ post:
         schema:
           oneOf:
             - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
+            - $ref: "../../builder-oapi.yaml#/components/schemas/Capella.SignedBlindedBeaconBlock"
```
ralexstokes, line 22: one thing to note is that providers of this endpoint won't be able to discriminate the type of the request without poking at the message type -- where usually we try to disambiguate the data based on some kind of "version" tag (see the response to this method, for example)

I compared this to the `beacon-apis` and the same thing is done there and I think conformity is more important here

One thing we can do to try to enhance this for providers of this endpoint is add an (optional) tag to the request header. We can follow the `beacon-apis` on this with the `Eth-Consensus-Version` HTTP header, as given here: https://github.com/ethereum/beacon-APIs/blob/093890f163788d9f78cbf0b8cd529e524cf6fe48/apis/beacon/blocks/blinded_blocks.yaml#L18

specs/builder.md
jimmygchen: I've separated capella changes into a separate folder `specs/capella`.

apis/builder/blinded_blocks.yaml
jimmygchen: Done! I agree - I raised it in the [4844 PR](https://github.com/ethereum/builder-specs/pull/61#discussion_r1066486846) but should've added it here.  `mev-boost` currently decodes with `capella` type and fallback to `bellatrix`, this could be prevented if consensus version is provided.

apis/builder/header.yaml
ralexstokes, line 48: I think we can roll back these changes now after using the new linter

builder-oapi.yaml
```diff
@@ -61,13 +61,21 @@ components:
       $ref: "./beacon-apis/types/bellatrix/execution_payload.yaml#/Bellatrix/ExecutionPayload"
     Bellatrix.SignedBlindedBeaconBlock:
       $ref: "./beacon-apis/types/bellatrix/block.yaml#/Bellatrix/SignedBlindedBeaconBlock"
-
     Bellatrix.SignedBuilderBid:
       $ref: "./types/bellatrix/bid.yaml#/Bellatrix/SignedBuilderBid"
+    Bellatrix.GetHeaderResponse:
+      $ref: "./types/bellatrix/bid.yaml#/Bellatrix/GetHeaderResponse"
```
ralexstokes, line 67: I think this is linked to the discriminator thing -- I don't see why we need the `GetHeaderResponse` here...

specs/bellatrix/validator.md
```diff
@@ -1,4 +1,4 @@
-# Builder -- Honest Validator
+# Bellatrix -- Builder -- Honest Validator
```
ralexstokes, line 1: nitpick
```suggestion
# Bellatrix -- Honest Validator
```

specs/capella/builder.md
```diff
@@ -0,0 +1,47 @@
...
+
+This is the modification of the builder specification accompanying the Capella upgrade.
+
+## Containers
+
+### Extended containers
+
+#### `ExecutionPayloadHeader`
+
+Note: `BuilderBid` and `SignedBuilderBid` types are updated indirectly.
```
ralexstokes, line 26: nice note

types/bellatrix/bid.yaml
```diff
@@ -21,3 +27,13 @@ Bellatrix:
...
+
+  GetHeaderResponse:
+    type: object
+    required:
+      - version
+    properties:
+      version:
+        type: string
+      data:
+        $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.SignedBuilderBid"
```
ralexstokes, line 39: think we can drop this

types/capella/bid.yaml
```diff
@@ -0,0 +1,30 @@
...
+  # Wrapper object in order to use `discriminator` with `oneOf`
+  GetHeaderResponse:
+    type: object
+    required:
+      - version
+    properties:
+      version:
+        type: string
+      data:
+        $ref: "../../builder-oapi.yaml#/components/schemas/Capella.SignedBuilderBid"
```
ralexstokes, line 30: ditto

ralexstokes (commented):
nice updates! I think the only thing left is to rollback the discriminator change as I think it is unnecessary given the linter update

apis/builder/header.yaml
jimmygchen: Thanks! reverted this change

builder-oapi.yaml
jimmygchen: Fixed

types/bellatrix/bid.yaml
jimmygchen: Fixed

types/capella/bid.yaml
jimmygchen: Fixed

apis/builder/header.yaml
jimmygchen: hmm I'm getting the same warning with the new linter :/ I'll have to take another look into this.

jimmygchen: Build passed with warnings here: https://github.com/ethereum/builder-specs/actions/runs/4099441364/jobs/7069352906

Same warnings on the `oneOf` validation + a few more weird ones that just appeared after switching the linter, I wonder if it's a bug in the linter, this issue is similar but we're getting this in linux/Mac as well https://github.com/Redocly/redocly-cli/issues/329

ralexstokes (approved):
spectacular! thanks for all of your work here :)

jimmygchen:
> spectacular! thanks for all of your work here :)

Thank you and everyone else for the review!

metachris:
For reference, here's the mev-boost issue to track Capella upgrades and further docs: https://github.com/flashbots/mev-boost/issues/451

metachris:
btw, thanks again @jimmygchen - awesome work 🙏

PR #61, opened 2023-01-03
Update getHeader and submitBlindedBlock to support blobs (EIP-4844)

jimmygchen:
Builder spec changes for EIP-4844, separated from PR #58. 

### `getHeader` changes

- Updated `ExecutionPayload` & `BeaconBlockBody`  types to align with EIP-4844 specs. (Depends on types from this Beacon API PR: https://github.com/ethereum/beacon-APIs/pull/271)
- Add `blinded_blobs` to `getHeader` response (`SignedBuilderBid`), so that validators can create signatures over the blinded block and blinded blobs.

### `submitBlindedBlock` changes

- Endpoint updated to accept `SignedBlindedBlockContents` after Deneb.
- Additionally returns unblinded blobs after Deneb; new type: `ExecutionPayloadAndBlobsBundle`.
- I've modified the response to return `blobs_bundle` instead of `blob_sidecars`, as I think it's more consistent with what we get from the EL, and BN should have enough to build the blob sidecars.

I'm currently pointing the `beacon-APIs` submodule to this PR, as it contains the new `SignedBlindedBlockContents` type: https://github.com/ethereum/beacon-APIs/pull/302

Moved design diagrams to [this HackMD page](https://hackmd.io/@jimmygchen/B1dLR74Io) as it was getting outdated here.


jimmygchen:
@ralexstokes @0xGabi would you mind reviewing this please? I've addressed all comments from the previous [PR](https://github.com/ethereum/builder-specs/pull/58). This PR contains changes for EIP-4844 only.

apis/builder/blinded_blocks.yaml
```diff
@@ -34,14 +40,20 @@ post:
...
                   - $ref: "../../builder-oapi.yaml#/components/schemas/Bellatrix.ExecutionPayload"
+                  - $ref: "../../builder-oapi.yaml#/components/schemas/Capella.ExecutionPayload"
+                  - $ref: "../../builder-oapi.yaml#/components/schemas/EIP4844.ExecutionPayloadAndBlobsSidecar"
           examples:
             bellatrix:
               $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.ExecutionPayload"
+            capella:
+              $ref: "../../builder-oapi.yaml#/components/examples/Capella.ExecutionPayload"
+            eip4844:
+              $ref: "../../builder-oapi.yaml#/components/examples/EIP4844.ExecutionPayloadAndBlobsSidecar"
```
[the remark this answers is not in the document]
jimmygchen: @g11tech there's been a bit of back and forth on this one in the previous PR and discord as well - originally I had similar thoughts about adding a new version, but I also agree that it's not necessary to add a new version to support new types if the fork information is provided in the request. 

Perhaps we could include the "consensus version" in the request headers (similar to the Beacon APIs), so it would be easier for the builder to support this? There's a [PR](https://github.com/flashbots/mev-boost/pull/431) in `mev-boost` to fallback to Bellatrix if it's not able to decode the payload, but I think this will get more complicated once the 4844 fork / payload is introduced.

To your point on code flow on the consensus side, I've tried implementing both in Lighthouse and haven't experienced too many issues with this (yet), but I don't have a strong opinion either way. The builder would need to know what version the payload is regardless of whether we do V1 or V2.

I think I'd go with the conventions (using version version without adding new api version) unless there's any strong reasons for adding v2? perhaps @ralexstokes can share some thoughts on this?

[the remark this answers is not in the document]
g11tech: as @ralexstokes pointed out, beacon apis insist on api version bump on format changes, i recon the response format has changed from direct execution payload to {execution_payload,...} object.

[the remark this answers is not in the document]
metachris: @g11tech not needing a new API versions for these changes was already discussed previously, i would keep it as is for now, and if you feel strongly, please initiate a new discussion about this in an issue.

jimmygchen:
I've replaced all EIP-4844 references with Deneb.
The PR for the rename in Beacon API is here: https://github.com/ethereum/beacon-APIs/pull/289

jimmygchen:
Waiting for the https://github.com/ethereum/beacon-APIs/pull/289 to be merged to pass the build

terencechain (commented):
As it stands today, blobs will be decoupled which changes how validator interacts with relayers (https://hackmd.io/cmYisgxkRuGe9NjX4gr97A) I'd recommend wait on it for a little bit until the spec matures

jimmygchen:
There is currently a discussion in the below PR regarding blob signing and publishing, and could have an impact on the builder API, I'll come back to this PR once we get more input on the decisison:

https://github.com/ethereum/beacon-APIs/pull/302#issuecomment-1454149870

jimmygchen:
@metachris @terencechain @g11tech @ralexstokes 
I've updated this PR to reflect some recent changes - there are quite a few changes to the types, so I've re-written the PR description. Please [see above](https://github.com/ethereum/builder-specs/pull/61#issue-1516822429).

Thanks!

specs/deneb/builder.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Containers
+
+### New containers
+
+#### `BlindedBlobsBundle`
+
+```python
+class BlindedBlobsBundle(Container):
+    commitments: List[KZGCommitment, MAX_BLOBS_PER_BLOCK]
```
jimmygchen, line 36: May have to update this to `MAX_BLOB_COMMITMENTS_PER_BLOCK`. 
https://github.com/ethereum/consensus-specs/pull/3338/files

specs/deneb/builder.md
```diff
@@ -0,0 +1,119 @@
...
+    graffiti: Bytes32
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    sync_aggregate: SyncAggregate
+    execution_payload_header: ExecutionPayloadHeader  # [Modified in Deneb]
+    bls_to_execution_changes: List[SignedBLSToExecutionChange, MAX_BLS_TO_EXECUTION_CHANGES]
+    blob_kzg_commitments: List[KZGCommitment, MAX_BLOBS_PER_BLOCK]  # [New in Deneb]
```
jimmygchen, line 118: May have to update `MAX_BLOBS_PER_BLOCK` to `MAX_BLOB_COMMITMENTS_PER_BLOCK`. 
https://github.com/ethereum/consensus-specs/pull/3338/files

metachris:
Thanks for your efforts @jimmygchen, much appreciated 🙏 

We've discussed this PR with the mev-boost maintainers group, and believe **a single `submitBlindedBlock` request would be preferable to requiring multiple API calls**. 
- Since all requests would be required to succeed anyway, it adds several downsides to split up the functionality over multiple API requests. 
- Not least that the relay is responsible for broadcasting the block _before_ it is allowed to send the payload back to the user. Having multiple API calls will require the relay to accumulate state internally, but at the same time block the response of other API calls until all are received, and the final block is validated, equivocation-checked and broadcast, and then there's still the remaining 1s delay before responding.

Also it may be a preferable approach to not return the payload to the proposer, but have it be the relay responsibility to broadcast the block. Returning the payload is currently delayed by relays, and is problematic due to equivocation attacks, and there has been some discussion recently by core devs about it perhaps being safer to not return the payload on this API.

cc/ @jtraglia @terencechain @ralexstokes @michaelneuder

jimmygchen:
@metachris the PR is updated to send a single `submitBlindedBlock`,  but diagram was outdated. It's now updated, thanks for pointing this out!

> Also it may be a preferable approach to not return the payload to the proposer, but have it be the relay responsibility to broadcast the block. Returning the payload is currently delayed by relays, and is problematic due to equivocation attacks, and there has been some discussion recently by core devs about it perhaps being safer to not return the payload on this API.

Thanks for sharing this - I'll keep an eye on the development!

examples/deneb/execution_payload_and_blobs_bundle.json
```diff
@@ -0,0 +1,41 @@
+{
+  "value": {
```
realbigsean, line 2: We don't actually have this top level `value` field right?

realbigsean (approved):

apis/builder/blinded_blocks.yaml
```diff
@@ -9,6 +9,9 @@ post:
 
     If the builder is not able to unblind the corresponding
     `ExecutionPayloadHeader`, it must error.
+
+    After Deneb, this endpoint will additionally accept `SignedBlindedBlobSidecars`, and return
+    unblinded blobs in response.
```
g11tech, line 14:
```suggestion
    with additional unblinded blobs in response.
```

examples/deneb/execution_payload_and_blobs_bundle.json
g11tech: +1

examples/deneb/execution_payload_and_blobs_bundle.json
```diff
@@ -0,0 +1,41 @@
...
+        "withdrawals": [
+          {
+            "index": "1",
+            "validator_index": "1",
+            "address": "0xabcf8e0d4e9587369b2301d0790347320302cc09",
+            "amount": "32000000000"
+          }
+        ]
+      },
+      "blobs_bundle": [
```
g11tech, line 32: do we want to keep it in line with execution-api's blobs_bundle? then it is an object or array items i.e. {blobs:[],commitments:[],proofs[]}

examples/deneb/execution_payload_and_blobs_bundle.json
```diff
@@ -0,0 +1,41 @@
...
+        {
+          "blob": "0x24564723180fcb3d994104538d351c8dcbde12d541676bb736cf678018ca4739",
+          "kzg_commitment": "0x8dab030c51e16e84be9caab84ee3d0b8bbec1db4a0e4de76439da8424d9b957370a10a78851f97e4b54d2ce1ab0d686f",
```
g11tech, line 35: i think we are moving away from prefixing this with `kzg_` to be future proof on notation and generality, same for proof

examples/deneb/signed_builder_bid.json
```diff
@@ -0,0 +1,35 @@
...
+          "timestamp": "1",
+          "extra_data": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+          "base_fee_per_gas": "1",
+          "excess_data_gas": "1",
+          "block_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+          "transactions_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+          "withdrawals_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
+        },
+        "blinded_blobs_bundle": {
+          "blob_roots": ["0x24564723180fcb3d994104538d351c8dcbde12d541676bb736cf678018ca4739"],
```
g11tech, line 25: blinded_blobs_bundle structure looks good :+1: , pls align blobs_bundle accordingly

g11tech (commented):
looks amazing overall :+1: , few suggestions

examples/deneb/execution_payload_and_blobs_bundle.json
jimmygchen: oh yes, it is what you suggested in the types for I forgot to update the example. Updated now. Thanks for pointing this out!

examples/deneb/execution_payload_and_blobs_bundle.json
jimmygchen: yep, same as above - I missed this example, fixed now. Thanks!

examples/deneb/execution_payload_and_blobs_bundle.json
jimmygchen: Oh I think this is the format `openapi lint` expects when validating the example files - we have them in all other example jsons, and is required for lint to work.

examples/deneb/signed_builder_bid.json
jimmygchen: Updated in https://github.com/ethereum/builder-specs/pull/61/commits/8e9aa72dcb1fdf5d66f712958ad4942c1ef25956

specs/deneb/builder.md
```diff
@@ -0,0 +1,119 @@
...
+#### `BuilderBid`
+
+Note: `SignedBuilderBid` is updated indirectly.
+
+```python
+class BuilderBid(Container):
+    header: ExecutionPayloadHeader # [Modified in Deneb]
+    value: uint256
+    pubkey: BLSPubkey
+    blinded_blobs: BlindedBlobsBundle  # [New in Deneb]
```
ralexstokes, line 94:
```suggestion
    blinded_blobs: BlindedBlobsBundle  # [New in Deneb]
    pubkey: BLSPubkey
```

this isn't written down anywhere but the idea is that when we have a signed message, the signature comes at the end of the message you sign over; and keeping with this ordering I think we should also put the `pubkey` here at the end of this message

ralexstokes (approved):
excellent work and I think this reflects the latest direction around 4844 well

I'd say we move ahead with merging this so we can start prototyping at various places in the mev-boost ecosystem/stack

specs/deneb/builder.md
jimmygchen: good to know - updated in https://github.com/ethereum/builder-specs/pull/61/commits/a97131c8d08a51eceab4c0ad9faf2d940d4e8b6e.

jimmygchen:
> excellent work and I think this reflects the latest direction around 4844 well
> 
> I'd say we move ahead with merging this so we can start prototyping at various places in the mev-boost ecosystem/stack

Thanks @ralexstokes ! I've updated the `beacon-API` submodule to latest (which contains the deneb types we need in this PR), so I think this should be ready to go from my POV.

There's a [pending PR](https://github.com/ethereum/consensus-specs/pull/3338/files) to update `MAX_BLOBS_PER_BLOCK` to `MAX_BLOB_COMMITMENTS_PER_BLOCK` in deneb `BeaconBlockBody` - I'm happy to update this when the other PR is finalised.

PR #64, opened 2023-01-25
Update linter to redocly

### LINKED ISSUE #63: Use `redocli` for CI linting
cf. https://github.com/ethereum/beacon-APIs/pull/223

jimmygchen:
Fixes #63.

Replace spectral with redocly, for the same reasons as https://github.com/ethereum/beacon-APIs/pull/223

We had some issues with the old linter in this PR, and @ralexstokes pointed out that we're using a different linter to the Beacon APIs.


.github/workflows/main.yml
```diff
@@ -11,11 +11,10 @@ jobs:
       with:
         submodules: true
     - uses: actions/setup-node@v2
-    - run: npm i -g @apidevtools/swagger-cli@4 @stoplight/spectral-cli@6.3.0
-    - run: spectral --version
+    - run: npm i -g @apidevtools/swagger-cli@4 @redocly/cli@1.0.0-beta.111
```
ralexstokes, line 14: have you looked at what `beta.111` provides over the default `1.0.0`?

.redocly.yml
```diff
@@ -0,0 +1,7 @@
+extends:
+  - recommended
+rules:
+  no-unused-components: error
+  operation-4xx-response: off
+  operation-tag-defined: error
+  security-defined: off
```
ralexstokes, line 7: do you know what this `security-defined` option does? it sounds important 😅

ralexstokes (commented):
looks great! and thanks for doing this :)

I left some questions but think we should merge soon

.github/workflows/main.yml
jimmygchen: The default version is the latest beta version - `1.0.0-beta.122` as of now:
https://www.npmjs.com/package/@redocly/cli?activeTab=versions

It looks like the version was pinned to `beta.111` in this PR due to a bug in the latest version of linter, but it wasn't clear to me what the issue was: 
https://github.com/ethereum/beacon-APIs/pull/268

I didn't spend too much time looking into this, but I've just raised a question on that PR - would be useful to know the reason / bug.

.redocly.yml
jimmygchen: 😅 Good question - I just assumed it would be easier / more consistent to use the exact same config as Beacon API. So it looks like if it's on, every operations will need to have security defined, e.g. JWT. Given all builder API endpoints don't require authentication, it would be easier to turn this off rather than adding `security: []` to every operation. No strong opinion on this, as long as we're consistent.

.github/workflows/main.yml
ralexstokes: looks like they responded -- should we also pin to the lower version?

.redocly.yml
ralexstokes: ah, ok sorry I thought this was an addition from the `beacon-apis` and I agree it is worth tracking consistency w/ the `beacon-apis`

thanks for looking into the option -- I agree this is the way!

.github/workflows/main.yml
ralexstokes: ah sorry, I see it is the same version, thanks for checking this out!

PR #73, opened 2023-04-11
Fixing path to builder and validator specs

fredriksvantes:
Since it was separated into different folders (bellatrix and capella) we can't link directly to a single file

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ with real tag name from step 1.):
 
 [ci]: https://github.com/ethereum/builder-specs/workflows/CI/badge.svg
 [oas-spec]: https://ethereum.github.io/builder-specs/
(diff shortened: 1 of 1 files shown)


ralexstokes (commented):
nice catch! I think we should just remove these links and drop the referents

I can fix this tomorrow or if you want to update the PR before I get to it then we can work from that

README.md
```diff
@@ -7,8 +7,6 @@ built by external entities.
 
 In this repository:
 * [API specification][oas-spec]
```
ralexstokes, line 9:
```suggestion
In this repository is the [API specification][oas-spec] along with specifications for actors in this ecosystem broken out by fork.
```

ralexstokes (approved):
lgtm, thanks for the update!

PR #76, opened 2023-04-25
add validations from API to spec

### LINKED ISSUE #48: Pull validations out of the API docs
There are some validations relevant for how implementers of the spec use the builder network that are "hidden" in the API documentation.

This organization is a bit different than how similar spec repos (like the `consensus-specs`) are styled so may lead to some confusion.

e.g. I suspect the root cause for this: https://github.com/flashbots/mev-boost/issues/245

Let's pull all validations out of the API docs (or at least put them into the "literate specs")

jacobkaufmann:
fixes #48

* define validations described in the builder API in the spec
* remove specific validity conditions from the API descriptions


specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+```
+
+## Constructing the `ExecutionPayloadHeader`
+
+Given the `slot`, `parent_hash`, and `pubkey`, the builder MUST return the header of the valid execution payload that is able to pay the `fee_recipient` for the registered `pubkey` the most.
+If the builder has not processed a previous validator registration for `pubkey`, then the builder MUST return (the appropriate error code).
+If the validator registered for `pubkey` is not the proposer for `slot`, then the builder MUST return (the appropriate error code).
+If `parent_hash` does not correspond to the head block hash according to the fork choice of the builder, then the builder MUST return (the appropriate error code).
+If possible under the rules of consensus, the builder MUST return an execution payload header whose `gas_limit` is equal to the `gas_limit` of the latest registration for `pubkey`.
+Otherwise, the builder MUST return an execution payload header with `gas_limit` as close as possible to the desired value under the rules of consensus.
```
jacobkaufmann, line 230: there is probably a more precise way to say this

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+```python
+def get_validator_with_pubkey(state: BeaconState, pubkey: BLSPubkey) -> Optional[Tuple[ValidatorIndex, Validator]]:
+    """
+    Look up validator for ``pubkey``
+    """
+
+    validator_pubkeys = [v.pubkey for v in state.validators]
+    if pubkey not in validator_pubkeys:
+        return None
```
ralexstokes, line 191: this is acceptable and I think a little more in line w/ the specs:

```suggestion
    validator_pubkeys = [v.pubkey for v in state.validators]
    assert pubkey in validator_pubkeys
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # Verify gas limit
+    # TODO: define `MAX_GAS_LIMIT` (?)
+    assert registration.gas_limit > 0 and registration.gas_limit < MAX_GAS_LIMIT
+
+    # Cache registration
+    latest_registrations[index] = registration
+```
+
+### `get_validator_with_pubkey`
```
ralexstokes, line 181: style: put "helpers" ahead of "core" logic, so here, `get_validator_with_pubkey` above `process_registration`

ralexstokes: same w/ other helpers below

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
 [above](#containers).
 
+## Registration processing
+
+### `process_registration`
+
+```python
+def process_registration(state: BeaconState,
+                         registration: SignedValidatorRegistrationV1,
+                         latest_registrations: Dict[ValidatorIndex, ValidatorRegistrationV1]):
```
ralexstokes, line 151: if we key this by public key and make the suggestion to `get_validator_with_pubkey` below we can just ignore needing to mess w/ the `index` at all

```suggestion
                         latest_registrations: Dict[BLSPublicKey, ValidatorRegistrationV1]):
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    Look up validator for ``pubkey``
+    """
+
+    validator_pubkeys = [v.pubkey for v in state.validators]
+    if pubkey not in validator_pubkeys:
+        return None
+
+    index = ValidatorIndex(validator_pubkeys.index(pubkey))
+    validator = state.validators[index]
+    return index, validator
```
ralexstokes, line 195: then just return the `validator`

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+```python
+def process_registration(state: BeaconState,
+                         registration: SignedValidatorRegistrationV1,
+                         latest_registrations: Dict[ValidatorIndex, ValidatorRegistrationV1]):
+    signature = registration.signature
+    registration = registration.message
+
+    # Verify BLS public key corresponds to a registered validator
+    validator = get_validator_with_pubkey(state, registration.pubkey)
+    assert validator is not None
```
ralexstokes, line 157: if we `assert` the validator exists in`get_validator_with_pubkey` then we don't need to do this here

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    # Verify validator registration elibility
+    index, validator = validator
+    assert is_eligible_for_registration(state, validator)
+
+    # Verify registration signature
+    assert verify_registration_signature(state, registration)
+
+    # If there exists a previous registration, then verify timestamp
+    # TODO: define `MAX_REGISTRATION_LOOKAHEAD`
+    prev_registration = latest_registrations[index]
```
ralexstokes, line 168: indexing this way will fail if the key is missing, we can use `get` instead
```suggestion
    prev_registration = latest_registrations.get(index)
```

or even just say:

```python
if index in latest_registrations:
    prev_registration = ...
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+### `is_eligible_for_registration`
+
+```python
+def is_eligible_for_registration(state: BeaconState, validator: Validator) -> bool:
+    """
+    Check if ``validator`` is not slashed and is either active or pending
+    """
+
+    if validator.slashed:
+        return False
```
ralexstokes, line 208: was this specified somewhere or is it being added here?

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    index = ValidatorIndex(validator_pubkeys.index(pubkey))
+    validator = state.validators[index]
+    return index, validator
+
+```
+
+### `is_eligible_for_registration`
+
+```python
+def is_eligible_for_registration(state: BeaconState, validator: Validator) -> bool:
```
ralexstokes, line 202: I interpret this condition as allowing any validator who is `active` or `pending` according to this document: https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ

we should make a pass on aligning conditions before merging this in

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    # If there exists a previous registration, then verify timestamp
+    # TODO: define `MAX_REGISTRATION_LOOKAHEAD`
+    prev_registration = latest_registrations[index]
+    if prev_registration is not None:
+        assert registration.timestamp > prev_registration.timestamp
+        assert registration.timestamp < (compute_timestamp_at_slot(state, state.slot) + MAX_REGISTRATION_LOOKAHEAD)
+
+    # Verify gas limit
+    # TODO: define `MAX_GAS_LIMIT` (?)
+    assert registration.gas_limit > 0 and registration.gas_limit < MAX_GAS_LIMIT
```
ralexstokes, line 175: I think we can skip this -- I don't think we need to be prescriptive about the gas limit beyond it being a `uint64`

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # Verify validator registration elibility
+    index, validator = validator
+    assert is_eligible_for_registration(state, validator)
+
+    # Verify registration signature
+    assert verify_registration_signature(state, registration)
+
+    # If there exists a previous registration, then verify timestamp
+    # TODO: define `MAX_REGISTRATION_LOOKAHEAD`
```
ralexstokes, line 167: 10 seconds, we can define a constants table like in the consensus-specs

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    prev_registration = latest_registrations[index]
+    if prev_registration is not None:
+        assert registration.timestamp > prev_registration.timestamp
```
ralexstokes, line 170:
```suggestion
        assert registration.timestamp >= prev_registration.timestamp
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+        assert registration.timestamp < (compute_timestamp_at_slot(state, state.slot) + MAX_REGISTRATION_LOOKAHEAD)
```
ralexstokes, line 171:
```suggestion
        assert registration.timestamp <= (compute_timestamp_at_slot(state, state.slot) + MAX_REGISTRATION_LOOKAHEAD)
```

ralexstokes: likely worth leaving a note that implementers of this API can work w/ wall clock time so could have a cut-off for this validation that is not aligned to a slot boundary

not sure we want to open the can of worms to include some kind of "clock" type here or not...

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+Otherwise, the builder MUST return an execution payload header with `gas_limit` as close as possible to the desired value under the rules of consensus.
+
+## Blinded block processing
+
+### `process_blinded_beacon_block`
+
+```python
+def process_blinded_beacon_block(state: BeaconState,
+                                 block: SignedBlindedBeaconBlock,
+                                 bids: Dict[Slot, Set[BuilderBid],
```
ralexstokes, line 239:
```suggestion
                                 bids: Dict[Slot, Set[BuilderBid]],
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    if prev_registration is not None:
+        assert registration.timestamp > prev_registration.timestamp
+        assert registration.timestamp < (compute_timestamp_at_slot(state, state.slot) + MAX_REGISTRATION_LOOKAHEAD)
+
+    # Verify gas limit
+    # TODO: define `MAX_GAS_LIMIT` (?)
+    assert registration.gas_limit > 0 and registration.gas_limit < MAX_GAS_LIMIT
+
+    # Cache registration
+    latest_registrations[index] = registration
```
ralexstokes, line 178: no need to cache this -- each fn should be pure wrt inputs

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+```python
+def process_blinded_beacon_block(state: BeaconState,
+                                 block: SignedBlindedBeaconBlock,
+                                 bids: Dict[Slot, Set[BuilderBid],
+                                 blocks: Dict[Slot, SignedBlindedBeaconBlock]):
+    # Verify block signature
+    assert verify_blinded_block_signature(state, block)
+
+    # TODO: Verify slot
```
ralexstokes, line 244:
```suggestion
    assert block.message.slot == state.slot
```

should work

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+### `process_blinded_beacon_block`
+
+```python
+def process_blinded_beacon_block(state: BeaconState,
+                                 block: SignedBlindedBeaconBlock,
+                                 bids: Dict[Slot, Set[BuilderBid],
+                                 blocks: Dict[Slot, SignedBlindedBeaconBlock]):
+    # Verify block signature
+    assert verify_blinded_block_signature(state, block)
```
ralexstokes, line 242: generally we want to put "heavy" checks after "light" checks so if there is some simple error we can bail before a node is pulled into doing a lot of work that would ultimately be unnecessary

signature verification will be one of the heavier things here so we should keep it near the end

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # TODO: Verify slot
+
+    block = block.message
+
+    # Verify a previous block has not been submitted for the same slot
+    assert blocks[block.slot] is None
```
ralexstokes, line 249: what's wrong w/ this?

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # Verify the execution payload header corresponds to a previous bid
+    bid_headers = [b.header for b in bids[block.slot]]
+    assert block.body.execution_payload_header in bid_headers
+
+    # TODO: Verify the remainder of the block
+
+    # Cache the block
+    blocks[block.slot] = block
```
ralexstokes, line 258: can drop this

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
 root and Builder API messages to the SSZ representations defined
 [above](#containers).
 
+## Registration processing
```
ralexstokes, line 144:
```suggestion
## Validator registration processing
```

specs/bellatrix/builder.md
```diff
@@ -141,6 +141,132 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+### `verify_registration_signature`
+
+```python
+def verify_registration_signature(state: BeaconState, signed_registration: SignedValidatorRegistrationV1):
+    pubkey = signed_registration.message.pubkey
+    signing_root = compute_signing_root(signed_registration.message, get_domain(state, DOMAIN_APPLICATION_BUILDER))
+    return bls.Verify(pubkey, signing_root, signed_registration.signature)
+```
+
+## Constructing the `ExecutionPayloadHeader`
```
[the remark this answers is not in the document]
ralexstokes: the builder should only be concerned w/ the context of `(slot, parent_hash, pubkey)`

ralexstokes, line 223: there's one step here "constructing the header", which I like that you broke out into its own thing, and another step which is "verifying the request to get the best bid/header"

let's keep each section separate

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,23 @@ the following actions:
    to assemble a `SignedBeaconBlock` following the rest of the proposal process outlined in the
    [Bellatrix specs][bellatrix-specs].
 
+##### `process_builder_bid`
+
+```python
+def process_builder_bid(state: BeaconState, bid: SignedBuilderBid):
+    # TODO: Verify bid signature
+    # TODO: Verify BLS public key corresponds to a registered validator (?)
```
ralexstokes, line 123: if im following how you are structuring this, it seems like this would be verifying the relay public key matches the one we expect

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,23 @@ the following actions:
...
+    # TODO: Verify bid value (?)
```
ralexstokes, line 124: we don't currently verify the bid value, so there wouldn't be a spec item for this

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,23 @@ the following actions:
...
+    # TODO: Verify execution payload header (?)
```
ralexstokes, line 125: the extent to which we do this is up for debate

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,23 @@ the following actions:
...
+```
+
+##### `prepare_blinded_beacon_block`
+
+```python
+def prepare_blinded_beacon_block(state: BeaconState, accepted_bid: BuilderBid) -> SignedBlindedBeaconBlock:
+    # TODO
```
ralexstokes, line 132: we can likely just call out to the CL specs for most of this

ralexstokes (commented):
awesome start! directionally looks great, I have some comments on lower-level things

and before writing more spec code I think we should align on exactly what validations should go into the spec

specs/bellatrix/builder.md
jacobkaufmann: if a previous, distinct `SignedBlindedBeaconBlock` was submitted for the same slot, then the validator is immediately slashable

specs/bellatrix/builder.md
jacobkaufmann: I ended up removing the helper method entirely, adopting the suggested pattern in `process_registration`

specs/bellatrix/builder.md
jacobkaufmann: see comment above w.r.t. refactor

specs/bellatrix/builder.md
jacobkaufmann: this was added, but I removed to remain consistent with existing "spec"

apis/builder/header.yaml
```diff
@@ -5,19 +5,10 @@ get:
...
-    of `parent_hash`, it must error. If `pubkey` does not match the builder's
-    expected proposer `pubkey` for `slot`, it must error. If the builder has
-    not received a validator registration associated with `pubkey`, it must
-    error.
-
-    When possible, the builder must return a header with `gas_limit` equal to
-    proposer's registered value. If this isn't possible due to the large
-    difference between `parent.gas_limit` and the preferred `gas_limit`, the
-    builder must return a header that moves `gas_limit` the maximum amount
-    allowed under the rules of consensus (currently `parent.gas_limit +/-
```
metachris, line 19: why remove this detail?

specs/bellatrix/builder.md
metachris: the relays accept registrations for active and pending validators

specs/bellatrix/builder.md
```diff
@@ -141,6 +147,103 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+```
+
+### `process_blinded_beacon_block`
+
+```python
+def process_blinded_beacon_block(state: BeaconState,
+                                 block: SignedBlindedBeaconBlock,
+                                 bids: Dict[Slot, Set[BuilderBid]],
+                                 blocks: Dict[Slot, SignedBlindedBeaconBlock]):
+    block = block.message
```
metachris, line 229: what's the goal of these tests here?

there's a ton of important checks in this call, maybe take a look at the relay implementation: https://github.com/flashbots/mev-boost-relay/blob/171c1aa8cf2d2252d3d7a7cf5aa378fdddf8bcfc/services/api/service.go#L999

apis/builder/header.yaml
jacobkaufmann: this is still WIP but the idea is to move specific validation criteria into the spec and outside of the API documentation

specs/bellatrix/builder.md
jacobkaufmann: the functions `is_eligible_for_activation` and `is_eligible_for_activation_queue` do not define the same "pending" criteria as the linked HackMD document. I'll modify the function to use the criteria defined there.

specs/bellatrix/builder.md
```diff
@@ -190,55 +202,141 @@ def process_registration(state: BeaconState,
...
+                                 bids: Dict[Slot, Set[ExecutionPayloadHeader]],
+                                 blocks: Dict[Slot, BlindedBeaconBlock]):
     block = block.message
 
-    # Verify slot
-    assert block.slot == state.slot
-
-    # Verify a previous block has not been submitted for the same slot
-    assert blocks[block.slot] is None
+    # Verify a previous, distinct block has not been submitted for the same slot
```
jacobkaufmann, line 334: call out the slashing somewhere above

specs/bellatrix/validator.md
jacobkaufmann: I removed this section for now

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,210 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    validator = state.validators[index]
+
+    # Verify validator registration elibility
+    assert is_eligible_for_registration(state, validator)
+
+    # If there exists a previous registration, then verify timestamp
+    if registration.pubkey in registrations:
+        prev_registration = registrations[registration.pubkey]
+        assert registration.timestamp >= prev_registration.timestamp
+        assert registration.timestamp <= (compute_timestamp_at_slot(state, state.slot) + MAX_REGISTRATION_LOOKAHEAD)
```
ralexstokes, line 208: a bit non-canonical but it will side-step this timing issue for now:

```suggestion
                         registrations: Dict[BLSPubkey, ValidatorRegistrationV1],
                         current_timestamp: int):
    signature = registration.signature
    registration = registration.message

    # Verify BLS public key corresponds to a registered validator
    validator_pubkeys = [v.pubkey for v in state.validators]
    assert pubkey in validator_pubkeys

    index = ValidatorIndex(validator_pubkeys.index(pubkey))
    validator = state.validators[index]

    # Verify validator registration elibility
    assert is_eligible_for_registration(state, validator)

    # If there exists a previous registration, then verify timestamp
    if registration.pubkey in registrations:
        prev_registration = registrations[registration.pubkey]
        assert registration.timestamp >= prev_registration.timestamp
        assert registration.timestamp <= current_timestamp + MAX_REGISTRATION_LOOKAHEAD
```

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,27 @@ the following actions:
...
+    domain = compute_domain(DOMAIN_APPLICATION_BUILDER)
+    signing_root = compute_signing_root(signed_registration.message, domain)
+    return bls.Verify(pubkey, signing_root, signed_bid.signature)
+```
+
+```python
+def process_bid(state: BeaconState, bid: SignedBuilderBid, parent_hash: Hash32, fee_recipient: ExecutionAddress):
+    # Verify execution payload header
+    header = bid.message.header
+    assert header.parent_hash == parent_hash
```
ralexstokes, line 132: let's just grab from the state for now

```suggestion
def process_bid(state: BeaconState, bid: SignedBuilderBid, fee_recipient: ExecutionAddress):
    # Verify execution payload header
    header = bid.message.header
    assert header.parent_hash == state.latest_execution_payload_header.block_hash
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,210 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    proposer = state.validators[signed_block.message.proposer_index]
+    epoch = compute_epoch_at_slot(signed_block.message.slot)
+    signing_root = compute_signing_root(signed_block.message, get_domain(state, DOMAIN_BEACON_PROPOSER, epoch))
+    return bls.Verify(proposer.pubkey, signing_root, signed_block.signature)
+```
+
+##### `process_blinded_beacon_block`
+
+```python
+def process_blinded_beacon_block(state: BeaconState,
```
ralexstokes, line 324: I think this fn is the weakest one here, if you are called to it you could add a comment that this is WIP

ralexstokes (commented):
excellent work! left a few comments but I think we should just merge this and move ahead

specs/bellatrix/builder.md
jacobkaufmann: I also realized that the check on the timestamp not being too far in the future should occur whether or not there is a previous registration, so I will fix that

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+To assist in registration processing, we use a function from the [consensus specs][consensus-specs]: [`is_active_validator`][is-active-validator].
+
+### `is_pending_validator`
+
+```python
+def is_pending_validator(validator: Validator, epoch: Epoch) -> bool:
+    """
+    Check if ``validator`` is pending in ``epoch``.
+    """
+    return validator.activation_epoch > epoch
```
ralexstokes, line 161: this will always be true?

if anything the check here seems to be "is this validator in the state"

ralexstokes: we check for membership below, so instead of checking for "is pending" I think we want to check they aren't exited (which also covers withdrawable)

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+```python
+def verify_registration_signature(state: BeaconState, signed_registration: SignedValidatorRegistrationV1):
+    pubkey = signed_registration.message.pubkey
+    domain = compute_domain(DOMAIN_APPLICATION_BUILDER)
+    signing_root = compute_signing_root(signed_registration.message, domain)
+    return bls.Verify(pubkey, signing_root, signed_registration.signature)
+```
+
+### `process_registration`
```
ralexstokes, line 185: I would add some language to tie the API docs to these docs

Even just adding a line here like:

`A `registration` is considered valid if the following function completes without raising any assertions:`

and similarly for the other endpoints getHeader and getPayload

open to whatever verbiage as long as the linkage is clear

the most we have in the API docs under this change is if something is "valid"

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+To assist in bidding, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`get_beacon_proposer_index`][get-beacon-proposer-index]
+* [`hash_tree_root`][hash-tree-root]
+
+Execution payloads are built for a specific `slot`, `parent_hash`, and `pubkey` tuple.
+
+The builder validates the tuple according to `is_eligible_for_bid(state, registrations, slot, parent_hash, pubkey)` where:
+
+* `registrations` is the registry of validators [successfully registered](#process-registration) with the builder
```
ralexstokes, line 235:
```suggestion
* `registrations` is the registry of validators [successfully registered](#process-registration) with the builder.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+To assist in bidding, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`get_beacon_proposer_index`][get-beacon-proposer-index]
+* [`hash_tree_root`][hash-tree-root]
+
+Execution payloads are built for a specific `slot`, `parent_hash`, and `pubkey` tuple.
+
+The builder validates the tuple according to `is_eligible_for_bid(state, registrations, slot, parent_hash, pubkey)` where:
+
```
ralexstokes, line 234:
```suggestion
* `state` is the builder's consensus state transitioned to `slot`, including the application of any blocks before `slot`.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+The validator accepts a bid and commits to a specific `ExecutionPayload` with a `SignedBlindedBeaconBlock`.
+
+### Bidding
+
+To assist in bidding, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`get_beacon_proposer_index`][get-beacon-proposer-index]
+* [`hash_tree_root`][hash-tree-root]
+
+Execution payloads are built for a specific `slot`, `parent_hash`, and `pubkey` tuple.
```
ralexstokes, line 231:
```suggestion
Execution payloads are built for a specific `slot`, `parent_hash`, and `pubkey` tuple corresponding to a unique beacon block serving as the parent.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+```python
+def is_eligible_for_bid(state: BeaconState,
+                        registrations: Dict[BLSPubkey, ValidatorRegistrationV1],
+                        slot: Slot,
+                        parent_hash: Hash32,
+                        pubkey: BLSPubkey) -> bool:
+    # Verify slot
+    if slot != state.slot:
+        return False
```
ralexstokes, line 245:
```suggestion
    assert slot == state.slot
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+### Bidding
+
+To assist in bidding, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`get_beacon_proposer_index`][get-beacon-proposer-index]
+* [`hash_tree_root`][hash-tree-root]
+
+Execution payloads are built for a specific `slot`, `parent_hash`, and `pubkey` tuple.
+
+The builder validates the tuple according to `is_eligible_for_bid(state, registrations, slot, parent_hash, pubkey)` where:
```
ralexstokes, line 233:
```suggestion
The builder validates requests for bids according to `is_eligible_for_bid(state, registrations, slot, parent_hash, pubkey)` where:
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+                        slot: Slot,
+                        parent_hash: Hash32,
+                        pubkey: BLSPubkey) -> bool:
+    # Verify slot
+    if slot != state.slot:
+        return False
+
+    # Verify BLS public key corresponds to a registered validator
+    if pubkey not in registrations:
+        return False
```
ralexstokes, line 249:
```suggestion
    assert pubkey in registrations
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # Verify BLS public key corresponds to the proposer for the slot
+    proposer_index = get_beacon_proposer_index(state)
+    if pubkey != state.validators[proposer_index].pubkey:
+        return False
```
ralexstokes, line 254:
```suggestion
    proposer_pubkey = state.validators[proposer_index].pubkey
    assert pubkey == proposer_pubkey
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+
+    # Verify parent hash
+    return parent_hash == state.latest_execution_payload_header.block_hash
```
ralexstokes, line 257:
```suggestion
    assert parent_hash == state.latest_execution_payload_header.block_hash
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+The builder validates the tuple according to `is_eligible_for_bid(state, registrations, slot, parent_hash, pubkey)` where:
+
+* `registrations` is the registry of validators [successfully registered](#process-registration) with the builder
+
+```python
+def is_eligible_for_bid(state: BeaconState,
+                        registrations: Dict[BLSPubkey, ValidatorRegistrationV1],
+                        slot: Slot,
+                        parent_hash: Hash32,
+                        pubkey: BLSPubkey) -> bool:
```
ralexstokes, line 242:
```suggestion
                        pubkey: BLSPubkey):
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
 root and Builder API messages to the SSZ representations defined
 [above](#containers).
 
+## Validator registration processing
+
```
ralexstokes, line 151:
```suggestion
Validators must send registrations before they can work with builders.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    # Verify registration signature
+    assert verify_registration_signature(state, registration)
+```
+
+## Building
+
+The builder builds execution payloads for registered validators.
+The builder submits a `SignedBuilderBid` that commits to the header of an execution payload.
+The builder only reveals the full execution payload once the validator accepts the bid.
+The validator accepts a bid and commits to a specific `ExecutionPayload` with a `SignedBlindedBeaconBlock`.
```
ralexstokes, line 222:
```suggestion
The builder builds execution payloads for registered validators and submits them to an auction that happens each slot.
When a validator goes to propose, they select the winning `SignedBuilderBid` offered in that slot
by constructing a `SignedBlindedBeaconBlock`.
The builder only reveals the full execution payload once they receive a valid `SignedBlindedBeaconBlock`.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    if pubkey != state.validators[proposer_index].pubkey:
+        return False
+
+    # Verify parent hash
+    return parent_hash == state.latest_execution_payload_header.block_hash
+```
+
+#### Constructing the `ExecutionPayloadHeader`
+
+Suppose the builder is bidding for `slot`, `parent_hash`, and `pubkey`.
```
ralexstokes, line 262:
```suggestion
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+The builder MUST submit a bid for the valid execution payload that is able to pay the `fee_recipient` for the registered `pubkey` the most.
```
ralexstokes, line 263:
```suggestion
The builder MUST provide a bid for the valid execution payload that is able to pay the `fee_recipient` in the validator registration for the registered `pubkey` the most.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+If possible under the rules of consensus, the builder MUST build an execution payload whose `gas_limit` is equal to the `gas_limit` of the latest registration for `pubkey`.
+Otherwise, the builder MUST build an execution payload with `gas_limit` as close as possible to the desired value under the rules of consensus.
```
ralexstokes, line 265:
```suggestion
The builder MUST build an execution payload whose `gas_limit` is equal to the `gas_limit` of the latest registration for `pubkey`, or as close as is possible under the consensus rules.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+### Revealing the `ExecutionPayload`
+
+#### Blinded block processing
+
+To assist in blinded block processing, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`compute_epoch_at_slot`][compute-epoch-at-slot]
+* [`get_domain`][get-domain]
+
+A proposer accepts a bid with a `SignedBlindedBeaconBlock`.
```
ralexstokes, line 310:
```suggestion
A proposer accepts a bid with a valid `SignedBlindedBeaconBlock`.
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    assert block.slot not in blocks or blocks[block.slot] == block
+
+    # Verify the execution payload header corresponds to a previous bid
+    assert block.slot in bids
+    assert block.body.execution_payload_header in bids[slot]
+
+    # Verify block signature
+    assert verify_blinded_block_signature(state, block)
+```
+
```
ralexstokes, line 345: let's just drop for now

```suggestion
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+To assist in blinded block processing, we use the following functions from the [consensus specs][consensus-specs]:
+
+* [`compute_epoch_at_slot`][compute-epoch-at-slot]
+* [`get_domain`][get-domain]
+
+A proposer accepts a bid with a `SignedBlindedBeaconBlock`.
+The proposer MUST accept at most one bid for a given `slot`.
+Otherwise, the builder can produce a [`ProposerSlashing`][proposer-slashing].
+
```
ralexstokes, line 313:
```suggestion

The builder must ensure the `SignedBlindedBeaconBlock` is valid according to the rules of consensus and also that the signature is correct for the expected proposer using `verify_blinded_block_signature`:

```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,215 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+##### `verify_blinded_block_signature`
+
+```python
+def verify_blinded_block_signature(state: BeaconState, signed_block: SignedBlindedBeaconBlock):
+    proposer = state.validators[signed_block.message.proposer_index]
+    epoch = compute_epoch_at_slot(signed_block.message.slot)
+    signing_root = compute_signing_root(signed_block.message, get_domain(state, DOMAIN_BEACON_PROPOSER, epoch))
+    return bls.Verify(proposer.pubkey, signing_root, signed_block.signature)
```
ralexstokes, line 321: I would actually get `proposer` from the `state` a la `get_beacon_proposer`

otherwise the spec would allow for the "honest behavior" where I get some kind of unbundling attack going on by sending a valid "sibling" block and capturing the mev into the canonical chain block

ralexstokes: prolly ditto on `epoch` just getting the current epoch

specs/bellatrix/validator.md
```diff
@@ -115,6 +115,29 @@ the following actions:
...
+##### Bid processing
+
+Bids received from step (1) above can be validated with `process_bid` below, where `state` corresponds to the state for the proposal without applying the block (currently under construction) and `fee_recipient` corresponds to the validator's most recently registered fee recipient address:
+
+```python
+def verify_bid_signature(state: BeaconState, signed_bid: SignedBuilderBid):
+    pubkey = signed_bid.message.pubkey
+    domain = compute_domain(DOMAIN_APPLICATION_BUILDER)
+    signing_root = compute_signing_root(signed_registration.message, domain)
+    return bls.Verify(pubkey, signing_root, signed_bid.signature)
```
ralexstokes, line 127:
```suggestion
    assert bls.Verify(pubkey, signing_root, signed_bid.signature)
```

specs/bellatrix/builder.md
jacobkaufmann: so this was added in based on your [previous comment](https://github.com/ethereum/builder-specs/pull/76#discussion_r1178448468). are you saying that we should only bother checking whether the validator is active? if that's incorrect, can you elaborate? thanks.

specs/bellatrix/builder.md
jacobkaufmann: since `is_eligible_for_bid` is a predicate, I was following the apparent convention in the consensus specs where, for example, [`is_eligible_for_activation`](https://github.com/ethereum/consensus-specs/blob/f1aabcd718eba26f83210b2545c9bfaa6531ea15/specs/phase0/beacon-chain.md#is_eligible_for_activation) returns a boolean instead of executing asserts. I don't have a strong preference here, but is there a reason to deviate from the consensus specs?

specs/bellatrix/validator.md
jacobkaufmann: similar to comment about predicates, I found that the consensus spec returns a boolean for verification, for example [`verify_block_signature`](https://github.com/ethereum/consensus-specs/blob/f1aabcd718eba26f83210b2545c9bfaa6531ea15/specs/phase0/beacon-chain.md#beacon-chain-state-transition-function) in the beacon state transition function. and similarly to other comment, I don't have a preference here so long as its consistent throughout the builder specs. if we keep the boolean instead of `assert`, then the function is missing the `-> bool` return type annotation.

specs/bellatrix/builder.md
jacobkaufmann: good look!

specs/bellatrix/builder.md
ralexstokes: the way this is defined `validator.activation_epoch > epoch` will just always be true as long as the validator is in the state (there should be no way to make a validator object where this is not true)

so it is kind of redundant if we do the membership check

that being said, it may be a bit more flexible if we do explicitly check and just leave this in

I would imagine the relays are just checking what ever status comes back from the API so to make it map more directly I'd favor just leaving this in for now

specs/bellatrix/builder.md
ralexstokes: how I read this is that `is_eligible_for_bid` is analogous to a top-level fn from the CL specs like `process_bid_request` where we do use the `assert` infra to signal an error (things like `is_eligible_for_activation` are "helpers" and in a sense subsidiary where we see the bool return value)

that being said I don't think it is super material here so you can do either one :)

specs/bellatrix/validator.md
ralexstokes: sure, thats fine then, we can match `verify_block_signature`

specs/bellatrix/builder.md
jacobkaufmann: okay cool. I will keep things as are for now, and we can reorganize later if desired

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,204 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+    Check if ``validator`` is active or pending.
+    """
+    epoch = get_current_epoch(state)
+    return is_active_validator(validator, epoch) or is_pending_validator(validator, epoch)
+```
+
+### `verify_registration_signature`
+
+```python
+def verify_registration_signature(state: BeaconState, signed_registration: SignedValidatorRegistrationV1):
```
ralexstokes, line 183:
```suggestion
def verify_registration_signature(state: BeaconState, signed_registration: SignedValidatorRegistrationV1) -> bool:
```

specs/bellatrix/builder.md
```diff
@@ -141,14 +147,204 @@ convert in-protocol messages to their SSZ representation to compute the signing
...
+A proposer selects a bid by constructing a valid `SignedBlindedBeaconBlock`.
+The proposer MUST accept at most one bid for a given `slot`.
+Otherwise, the builder can produce a [`ProposerSlashing`][proposer-slashing].
+
+The builder must ensure the `SignedBlindedBeaconBlock` is valid according to the rules of consensus and also that the signature is correct for the expected proposer using `verify_blinded_block_signature`:
+
+##### `verify_blinded_block_signature`
+
+```python
+def verify_blinded_block_signature(state: BeaconState, signed_block: SignedBlindedBeaconBlock):
```
ralexstokes, line 327:
```suggestion
def verify_blinded_block_signature(state: BeaconState, signed_block: SignedBlindedBeaconBlock) -> bool:
```

ralexstokes (approved):
looks good!

for those following along, this PR is a step to massage the builder-specs to a place where we can generate test vectors from them to assist in testing users of the specs/APIs

something that has become apparent while working on this PR is the "builder" abstraction that papers over the differences b/t the relay and builder actors; I think we should continue to refine this spec so this distinction is clearer (and some of the logic in this spec would move to the relay specs)

PR #78, opened 2023-05-21
add initial scaffolding for pyspec

### LINKED ISSUE #77: add scaffolding for executable python spec
the ethereum consensus specs are executable, such that the specs are testable, and test vectors can be generated for consensus clients to consume. the builder specs should be executable in a similar way.

the scope of this issue is to set up the basic scaffolding for executable builder specs.

for example:

* script to parse the markdown files and generate python files with defined constants, containers, functions, etc
* minimal test(s)
* mechanism to execute the test(s) (e.g. `Makefile`)

jacobkaufmann:
resolves #77

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+**/__pycache__/
+venv
+
+tests/core/pyspec/builderspec/bellatrix.py
+tests/core/pyspec/builderspec/capella.py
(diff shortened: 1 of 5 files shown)


jacobkaufmann:
still a WIP

PR #79, opened 2023-05-29
Update block's `blob_kzg_commitments` size limit

jimmygchen:
Small change to update block's `blob_kzg_commitments` size limit.
Consensus spec PR merged: https://github.com/ethereum/consensus-specs/pull/3338.


ralexstokes (approved):
ty!

PR #80, opened 2023-05-29
Fix incorrect field name in Deneb `BuilderBid` type

jimmygchen:
I've just noticed that the field names in the `BuilderBid` schema doesn't match the `SignedBuilderBid` example JSON:

https://github.com/ethereum/builder-specs/blob/0d039df6cbe8af9ff16a1546281064d41638b5e6/examples/deneb/signed_builder_bid.json#L24-L28

- I've updated the field name to `blinded_blobs_bundle` to match the example. (I think this name also reflects the structure better)
- Fixed the ordering of the fields in the example to match the schema, i.e. `blob_roots` comes last in `blinded_blobs_bundle`


ralexstokes (approved):
excellent, thank you!

PR #84, opened 2023-07-27
Rename "data gas" to "blob gas" in deneb examples

StefanBratanov:
Address https://github.com/ethereum/EIPs/pull/7354


ralexstokes (approved):
thank you!

PR #85, opened 2023-07-27
Update maxItems from 4 to 6 in BlobsBundle

jtraglia:
This should be updated from 4 to 6 to match `MAX_BLOBS_PER_BLOCK` in the spec.


ralexstokes (approved):
great thx!

PR #86, opened 2023-08-03
Update beacon api spec to latest changes

avalonche:


ralexstokes (approved):
not exactly sure where this updates the `beacon-apis` submodule but directionally it looks good!

Issue #53, opened 2022-10-07
Allow SSZ encoding in addition to JSON encoding for all the requests


metachris:
I've benchmarked encoding and decoding of signed-blinded-beacon-block payloads with SSZ vs JSON: https://github.com/flashbots/go-boost-utils/pull/50

SSZ seems 40-50x faster.

Each `getPayload` call currently does 8 JSON encoding and decoding steps (if we include mev-boost in the mix):

1. BN encodes the `getPayload` request body (signed blinded beacon block)
2. mev-boost decodes the `getPayload` request body
3. mev-boost encodes the `getPayload` request body
4. relay decodes the `getPayload` request body
5. relay encodes the `getPayload` response body (execution payload) 
6. mev-boost decodes the `getPayload` response body
7. mev-boost encodes the `getPayload` response body
8. BN decodes the `getPayload` response body

Considering 20-40ms per coding on average, that's up to 200-300ms JSON latency (or more). 

Sending the data SSZ encoded could reliably shave 200-250ms off each `getPayload` roundtrip.

ralexstokes:
in general I think this would be great to add ASAP!

> How can this work with relays that didn't (yet) support SSZ encoded payloads?

can always just make a `v2` of the API that supports either codec via the header (this is how the beacon-apis do it) and leave the `v1` as-is

otherwise, we update the `v1` to have current behavior w/ no header, and then also add support for `Content-Type: application/json` or `Content-Type: application/ssz` based on what the user wants -- if a caller requests something a relay cannot provide they just respond with a HTTP 400 error

metachris:
Thinking about it, upgrading `v1` to allow for it seems kind of straight forward this way: 

1. beacon-node calls `getHeader` with request header `accept-encoding: application/ssz`
   1. If response is SSZ encoded, then it uses SSZ for `getPayload`
   2. If response is JSON encoded, then use JSON for `getPayload`

Should be easy to notify other release before this change gets out, for them to add least make sure they can handle it correctly and not error out (even if not using SSZ).

metachris:
cc/ @terencechain @tbenr @StefanBratanov @tersec @realbigsean @dapplion @lightclient

dapplion:
@metachris Could you expand the benchmarks to different block sizes within what's observed in mainnet?

Agree that would be great to add this asap. No strong opinions on whether to bump to v2 or not

tersec:
Agree with @dapplion that this would be a good addition.

[...]

PR #90, opened 2023-10-30
Drop "blinded blob sidecar" concept after using signed header to authenticate blobs

ralexstokes:
Following https://github.com/ethereum/consensus-specs/pull/3531, we can make some strong simplifications to the builder-specs which this PR implements.

Changes to `getHeader`:
The relay must still provide `KZGCommitments` in a signed bid but they no longer need to pass along proofs or blob roots.

Changes to `getPayload`:
The relay simply needs the `SignedBlindedBeaconBlock` (as per the status quo) and can still assemble the unblinded beacon block. The relay will still reveal the execution payloads alongside the full "blob bundle".

Open questions:
The simplification from the linked `consensus-specs` PR requires blob gossip to include a short Merkle proof from the signed header's body root to each KZG commitment it includes. This PR assumes the relay will do this work independently although we could demand the proposer sends these along as well with the `SignedBlindedBeaconBlock`.

Pros: less computation for relays (and less code to debug/maintain)
Cons: demands consensus clients have done this work ahead of time (they may wait until after selecting the remote payload as the blob sidecar gossip can come after)


ralexstokes:
> As the relay will need to verify proposer inputs anyways it seems like relays will need to do the same amount of compute with more work on the consensus clients.

there is a small optimization here where the beacon node (who can optimistically compute the inclusion proofs) can send them to the relay, so that the relay only has to do _verification_ (and not _compute_ and _verify_ themselves)

the question comes down to: are relays ok to maintain this code to compute the proofs? or should they rely on beacon nodes to compute and send along with the signed header?

apis/builder/blinded_blocks.yaml
```diff
@@ -22,22 +21,22 @@ post:
...
-            - $ref: "../../builder-oapi.yaml#/components/schemas/Deneb.SignedBlindedBlockContents"
+            - $ref: "../../builder-oapi.yaml#/components/schemas/Deneb.SignedBlindedBeaconBlock"
         examples:
           bellatrix:
             $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBlindedBeaconBlock"
           capella:
             $ref: "../../builder-oapi.yaml#/components/examples/Capella.SignedBlindedBeaconBlock"
           deneb:
-            $ref: "../../builder-oapi.yaml#/components/examples/Deneb.SignedBlindedBlockContents"
+            $ref: "../../builder-oapi.yaml#/components/examples/Deneb.SignedBlindedBeaconBlock"
```
g11tech, line 39: as discussed in call would prefer the proofs be submitted along

ralexstokes: just pushed the update to require proofs from proposer

ralexstokes:
linter is failing as it depends on https://github.com/ethereum/beacon-APIs/pull/369, will update once that PR is merged

g11tech (approved):
lgtm

potuz:
I'd like to get a little more discussion about this design. I can see clear tradeoffs between the proposer doing it or the relay doing it.  On the one hand this reduces code complexity on the relay and adds it to the proposer that presumably is running a more vetted software. On the other hand it adds latency to block production as now the proposer needs to send back the proofs. The relay still needs to assemble the sidecar with the actual blob contents. If we were to have the relay assembling the sidecars beforehand, it can start building most of the proof (out of the 17 entries in the proof, only the last four require entries from the Beacon Block Body). So most of the work can be done before the blinded beacon block comes back. 

Even if we were to force the proposer to send the proofs, an intermediate step that would alleviate the extra data on the wire is to send just the last four roots in the proof and let the relay compute the remaining 13. 

Given the above it seems to me that it's more beneficial for the network as a whole to have the relay building the proofs and not the proposers.

realbigsean:
Can the proposer start computing the proofs and cache them while the validator is signing the block?

potuz:
The proposer can, but the problem is sending them over the wire. These are 3264 bytes for 6 blobs

ralexstokes:
@potuz this is a valid solution but it does complicate things even more than "proposer sends full proofs" or "relay and proposer both compute them independently"

im not convinced the additional latency is going to matter that much as proof creation should be quite cheap and like @realbigsean is saying either or both parties can pipeline proof creation as an optimization

I would need to see a prototype demonstrating the latency gain is enough to warrant the more complex (especially conceptually) solution you are proposing. The latest change here has the proposer compute the proofs in fact to make things as simple as possible for the relay.

potuz:
What I'm proposing is that the relay does the full proof so that the proposer doesn't incurr in any extra network latency. The argument about the majority of the proof being computable by the relay before getting any signature from the proposer is just to exemplify this. If instead of counting the 17 roots you count only those 13, the proposer will be sending 2496 extra bytes that could have been completely computed by the relay before even sending the bid

realbigsean:
26,112 bits / 10 Mbps is about 2.6ms, since we'd be sending it back and forth it'd be more like >5ms total additional latency. Seems like it's not insignificant, so this is kind of making me think the relay should handle just this.

realbigsean:
26,112 bits / 10Mbps = 2.6ms . That on top of the optimization @potuz is suggesting seems significant. It also does make the API a bit simpler. As far as relay complexity, all clients will have code for the proof computation implemented so the methods should be pretty widely available. It gets a bit complex if they actually implements potuz optimization. but overall it does seem worth doing. Are we able to get feedback from people actually running relays?

g11tech:
not in favor of partial optimizations to save a few ms over http, so latency isn't a factor here imo and we should  either let all proofs come from beacon or all proofs generated by the relay

i prefer beacon but again not so hard bend on this.

avalonche:
agree that the optimisations of a few ms doesn't seem significant as there are better optimisations available (like [this issue](https://github.com/ethereum/builder-specs/issues/53) with 200-300ms improvements) and proofs should either all be computed on the relay or from the beacon node. 
Are there examples / libraries on how to implement the optimisation @potuz mentioned? From the perspective of relay code complexity it shouldn't be too much of an issue to include proof creation but not the more complex optimisation proposed.

specs/deneb/builder.md
```diff
@@ -29,55 +31,33 @@ This is the modification of the builder specification accompanying the Deneb upg
...
-#### `BlindedBlobsBundle`
+#### `BlobsBundle`
 
-```python
-class BlindedBlobsBundle(Container):
-    commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
-    proofs: List[KZGProof, MAX_BLOB_COMMITMENTS_PER_BLOCK]
-    blob_roots: List[Root, MAX_BLOB_COMMITMENTS_PER_BLOCK]
-```
+Same as [`BlobsBundle`](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/validator.md#blobsbundle) in Deneb consensus specs.
```
etan-status, line 36: Needs added `List[...]` instead of `Sequence` though, to enable SSZ transport.

`MAX_BLOB_COMMITMENTS_PER_BLOCK` limit is fine.

specs/deneb/builder.md
```diff
@@ -29,55 +31,33 @@ This is the modification of the builder specification accompanying the Deneb upg
...
-#### `ExecutionPayloadAndBlobsBundle`
+#### `SignedBlindedBeaconBlockWithKZGCommitmentInclusionProofs`
 
 ```python
-class ExecutionPayloadAndBlobsBundle(Container):
-    execution_payload: ExecutionPayload
-    blobs_bundle: BlobsBundle
+class SignedBlindedBeaconBlockWithKZGCommitmentInclusionProofs(Container):
+    signed_blinded_beacon_block: SignedBlindedBeaconBlock
+    kzg_commitment_inclusion_proofs: List[KZGCommitmentInclusionProof, MAX_BLOB_COMMITMENTS_PER_BLOCK]
```
etan-status, line 60: Could be compacted by putting the `kzg_commitments` instead.

The caller can then compute the `kzg_commitment_inclusion_proofs` from `kzg_commitments` + `signed_blinded_beacon_block`. The proofs contain redundant data that can also be found in the `SignedBlindedBecaonBlock`.

etan-status: actually, `SignedBlindedBeaconBlock` already contains `kzg_commitments`, so it is pointless to also include `kzg_commitment_inclusion_proofs` as it is fully recoverable from just the `SignedBlindedBeaconBlock`.

types/deneb/block.yaml
```diff
@@ -0,0 +1,12 @@
...
+    type: object
+    properties:
+      signed_blinded_beacon_block:
+        $ref: "../../beacon-apis/types/deneb/block.yaml#/Deneb/SignedBlindedBeaconBlock"
+      kzg_commitment_inclusion_proofs:
+        type: array
+        items:
+          $ref: '../../beacon-apis/types/deneb/blob_sidecar.yaml#/KZGCommitmentInclusionProof'
+        minItems: 0
+        maxItems: 6
```
etan-status, line 12: the `maxItems` should be the theoretical max: `MAX_BLOB_COMMITMENTS_PER_BLOCK`, for stable SSZ transfer

potuz:
> agree that the optimisations of a few ms doesn't seem significant as there are better optimisations available 

It doesn't make much sense for the proposer to get the kzg commitments, then compute the proofs and then send them back when the relay has everything it needs to start computing this before it even sends the list to the proposer. The design is simply much cleaner if the relay is in charge of this. 

It is also hard for the proposer to parallelize this since at the same time it will be computing proofs for its local block (together with all the other tasks when proposing). And as a plus is better forward looking for ePBS when the kzg commitments won't even be in the block and the builder will have to include these proofs in the payload.

ralexstokes:
yeah i think this is reasonable, i'll roll back to the original state with independent computation later today 

and we will be sure to call out the relevant code for the relays and make sure they can pass test vectors

types/deneb/block.yaml
ralexstokes: going to have relay compute independently, so not a concern anymore

ralexstokes:
@avalonche @g11tech @potuz @realbigsean @etan-status 

I've made the relevant updates and think this is ready to merge in the next day or two

I'll follow back up with some pointers on how relays should compute and verify KZG commitment inclusion proofs in another PR

g11tech:
Considering ePBS then it makes sense for relay to do the  compute 💪

potuz:
Can't approve it cause I'm out of my depth with regards to APIs, but this looks good to me!

types/deneb/blobs_bundle.yaml
```diff
@@ -15,20 +15,6 @@ Deneb:
...
-        type: array
-        items:
-          $ref: "../../beacon-apis/types/primitive.yaml#/Root"
-        minItems: 0
-        maxItems: 6
-
-
   BlobsBundle:
     allOf:
       - $ref: '#/Deneb/BlobsBundleCommon'
```
g11tech, line 20: small cleanup: i think `BlobsBundleCommon` would not be required, just `BlobsBundle`

g11tech (approved):
lgtm

types/deneb/bid.yaml
```diff
@@ -7,9 +7,13 @@ Deneb:
...
             description: "`ExecutionPayloadHeader` to use in block proposal."
-          blinded_blobs_bundle:
-            $ref: "./blobs_bundle.yaml#/Deneb/BlindedBlobsBundle"
-            description: "`BlindedBlobsBundle` to use in block proposal."
+          blob_kzg_commitments:
+            type: array
+            items:
+              $ref: '../../beacon-apis/types/primitive.yaml#/KZGCommitment'
+            minItems: 0
+            maxItems: 6
```
realbigsean, line 15: I think we want this to be `MAX_BLOB_COMMITMENTS_PER_BLOCK` (4096)

ralexstokes: ah good catch!

PR #98, opened 2024-01-20
Add required properties to object types

nflaig:
- See https://github.com/ethereum/beacon-APIs/pull/399 for motivation / rationale

Note that there is no need to add required field to request bodies as this is already set correctly


ralexstokes (approved):
looks good, thanks!

PR #100, opened 2024-04-29
Replace oneOf with anyOf in openapi schemas

nflaig:
Current openapi spec is invalid, see [validation result](https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fethereum.github.io%2Fbuilder-specs%2Freleases%2Fv0.4.0%2Fbuilder-oapi.json).

See https://github.com/ethereum/beacon-APIs/pull/421 for more rational.

### DIFF
--- a/apis/builder/blinded_blocks.yaml
+++ b/apis/builder/blinded_blocks.yaml
@@ -26,7 +26,7 @@ post:
     content:
       application/json:
         schema:
-          oneOf:
+          anyOf:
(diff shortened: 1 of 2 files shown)


nflaig:
would be good to get this in before the next builder spec release, this makes running spec tests more difficult as api spec is invalid and requires workarounds

ralexstokes (approved):
interesting, if this works how I think it does then it makes sense!

thanks

PR #101, opened 2024-05-01
add initial electra spec

jacobkaufmann:
add initial `electra` spec.

## NOTES
 

relevant PRs:
* https://github.com/ethereum/execution-apis/pull/587
* https://github.com/ethereum/beacon-APIs/pull/469

### DIFF
--- a/specs/electra/builder.md
+++ b/specs/electra/builder.md
@@ -1,0 +1,54 @@
+# Electra -- Builder Specification
+
+## Introduction
+
+This is the modification of the builder specification accompanying the Electra upgrade.
+
+## Containers
+
+### Extended containers
+
+#### `ExecutionPayloadAndBlobsBundle`
+
+```python
+class ExecutionPayloadAndBlobsBundle(Container):
+    execution_payload: ExecutionPayload # [Modified in Electra]
+    blobs_bundle: BlobsBundle
+```
+
+#### `BuilderBid`
+
+Note: `SignedBuilderBid` is updated indirectly.
+
+```python
+class BuilderBid(Container):
+    header: ExecutionPayloadHeader # [Modified in Electra]
+    blob_kzg_commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    value: uint256
+    pubkey: BLSPubkey
+```
+
+#### `ExecutionPayloadHeader`
+
+See [`ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#executionpayloadheader) in Electra consensus specs.
+
+##### `BlindedBeaconBlockBody`
+
+Note: `BlindedBeaconBlock` and `SignedBlindedBeaconBlock` types are updated indirectly.
+
+```python
+class BlindedBeaconBlockBody(Container):
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data
+    graffiti: Bytes32
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    sync_aggregate: SyncAggregate
+    execution_payload_header: ExecutionPayloadHeader  # [Modified in Electra]
+    bls_to_execution_changes: List[SignedBLSToExecutionChange, MAX_BLS_TO_EXECUTION_CHANGES]
+    blob_kzg_commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    consolidations: List[SignedConsolidation, MAX_CONSOLIDATIONS]  # [New in Electra]
+```
--- a/specs/electra/builder.md
+++ b/specs/electra/builder.md
@@ -8,7 +8,7 @@ This is the modification of the builder specification accompanying the Electra u
 
 ### Extended containers
 
-##### `BlindedBeaconBlockBody`
+#### `BlindedBeaconBlockBody`
 
 Note: `BlindedBeaconBlock` and `SignedBlindedBeaconBlock` types are updated indirectly.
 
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
looks good! may hold off on merging until after the first iteration of pectra following devnet-0

terencechain (approved):

specs/electra/builder.md
```diff
@@ -0,0 +1,54 @@
...
+## Introduction
+
+This is the modification of the builder specification accompanying the Electra upgrade.
+
+## Containers
+
+### New containers
+
+#### `ExecutionBundle`
+
```
ralexstokes, line 12: add note about "this replaces the `ExecutionPayloadAndBlobsBundle`

specs/electra/builder.md
```diff
@@ -0,0 +1,54 @@
+# Electra -- Builder Specification
+
+## Introduction
+
+This is the modification of the builder specification accompanying the Electra upgrade.
+
```
ralexstokes, line 6: another line of text: "spec works like it did in previous forks, with changes to types as given below"

ralexstokes (changes requested):
looks good!

we are blocked on the `beacon-apis` but this should be good enough to get implementer's started

specs/electra/builder.md
```diff
@@ -0,0 +1,60 @@
...
+#### `ExecutionBundle`
+
+The `ExecutionBundle` supersedes the [`ExecutionPayloadAndBlobsBundle`][execution-payload-and-blobs-bundle-deneb].
+
+```python
+class ExecutionBundle(Container):
+    execution_payload: ExecutionPayload
+    blobs_bundle: BlobsBundle
+    execution_requests: ExecutionRequests
```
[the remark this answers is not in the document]
nflaig: No longer the case since https://github.com/ethereum/consensus-specs/pull/3875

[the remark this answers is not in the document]
rkrasiuk: is anybody working on updating `execution-apis`?

ralexstokes (approved):
updates lgtm, thanks!

lucassaldanha (approved):

specs/electra/builder.md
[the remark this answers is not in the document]
lucassaldanha: Yes. Here is the WIP: https://github.com/ethereum/execution-apis/pull/587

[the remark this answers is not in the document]
lucassaldanha: Oh just noticed that you did it... 👍 
Nevermind :)

specs/electra/builder.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### `BuilderBid`
+
+Note: `SignedBuilderBid` is updated indirectly.
+
+```python
+class BuilderBid(Container):
+    header: ExecutionPayloadHeader
+    blob_kzg_commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    execution_requests_root: Root # [New in Electra]
```
lucassaldanha, line 34: Unless I am missing something, the CL won't be able to validate the block if we don't have the actual `ExecutionRequests` in the `BuilderBid`. Instead of the `execution_requests_root`, we want the actual `execution_requests` here.

potuz: Interestingly enough, the root is enough in the blinded block part when the message returns to the relay, so at least there' no need for a double trip of the full requests.

mkalinin: @potuz do you mean they have to be in the `BuilderBid` but not necessarily in the `SignedBlindedBeaconBlock`?

jacobkaufmann: we have the root of the execution requests in the `BuilderBid` because those are observable, and there are concerns about MEV leakage if the execution requests are not hidden.

it is okay to only have the root of the execution requests in the `BuilderBid`, because the proposer can construct the `SignedBlindedBeaconBlock` with only the root.

once the relay receives and verifies the `SignedBlindedBeaconBlock`, it can broadcast the `SignedBeaconBlock`, revealing both the execution payload and the execution requests.

terencechain: `BlindedBeaconBlockBody` needs to contain `excution_requests` instead of `excution_requests_root`
Proposer needs the full execution requests, or else it can not calculate the beacon state root to sign over

jacobkaufmann: I see. we will need to refactor. good callout @terencechain!

lucassaldanha (approved):
LGMT

specs/electra/builder.md
terencechain, line 21: I dont think you need `execution_requests` here. `execution_requests` is already part of `BuilderBid`

ensi321: > I dont think you need `execution_requests` here. `execution_requests` is already part of `BuilderBid`

In this case we don't need `ExecutionBundle` at all, just keep `ExecutionPayloadAndBlobsBundle`

nflaig (approved):

specs/electra/builder.md
```diff
@@ -0,0 +1,47 @@
...
+
+#### `BuilderBid`
+
+Note: `SignedBuilderBid` is updated indirectly.
+
+```python
+class BuilderBid(Container):
+    header: ExecutionPayloadHeader
+    blob_kzg_commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    execution_requests: ExecutionRequests # [New in Electra]
```
mkalinin, line 21: Should this field type be switched to `bytes` If https://github.com/ethereum/execution-apis/pull/591 is accepted?

lucassaldanha: I think it makes sense to keep them consistent. So it makes sense to change it. Otherwise we need two decoding implementations on the CL side.

ralexstokes: I think I would leave this as-is in the spec

can mirror the Engine API for the Builder API HTTP definition

ensi321 (approved):

specs/electra/builder.md
ralexstokes: to @lucassaldanha point, given that this version will be in the `beacon-apis`, I don't think we are making an additional ask of CLs to deal with this version here

will go ahead and merge this for now to unblock devnet-4, if we decide we want to mirror the Engine API we can address in a future PR

nflaig: > Should this field type be switched to bytes

What would be the difference of this compared to using SSZ encoding for the whole `BuilderBid`? It seems much cleaner to have the option to use JSON (for debugging purposes) or SSZ in transit and is what we do on the beacon-api for most of the time sensitive apis.

If we are concerned about latency, should move to using SSZ on the builder api as proposed in https://github.com/ethereum/builder-specs/pull/104

ralexstokes: yeah I think that's a nice resolution here! and honestly likely how we should be thinking about it in the Engine API but full SSZ there will take more time

nflaig: > but full SSZ there will take more time

yeah I see why this is done on the engine api, but please not on CL side, embedding ssz inside json data is just the worst of both worlds, it's harder to debug issues during development and you still send json around for all others fields + it does not allow streaming the response. Ideally, we only wanna use json for development and stuff like kurtosis testing.

PR #54, opened 2022-10-10
[wip] SSZ encoding for getHeader and getPayload

metachris:
Allows SSZ encoded payloads for getHeader and getPayload: https://github.com/ethereum/builder-specs/issues/53

This PR adds:
1. `getHeader` header: `Accept`
2. `getHeader` response encoding: `application/octet-stream`
3. `getPayload` request and response encoding: `application/octet-stream`

PR #104, opened 2024-07-09
Add SSZ support to builder api

nflaig:
This is a proposal to add SSZ support to builder API in a backward compatible way without requiring to bump the apis to v2.

It outlines what is expected from the client and how a well-behaved server should handle different content types, headers, and use appropriate error codes to allow the client to handle the error.

Notable changes are
- `submitBlindedBlock` allows to SSZ encoding for both request and response body
- `getHeader` allows SSZ encoding of response body
- Define error codes for unsupported media types (406, 415), those are not mandatory but it would be good if servers adopt these over time to allow clients to implement specific error handling
- Adds `Eth-Consensus-Version` to responses, only required if response body is SSZ encoded but preferably should be added to all responses eventually
- Add notes on how a well-behaved client and server implementation should work

Since we can't know if a builder implemented the changes as per spec, the client can determine if SSZ is supported by requesting an SSZ encoded response in the `getHeader` request. Preferably this should be done setting a `Accept` header with multiple entries and quality values like `Accept: application/octet-stream;q=1.0,application/json;q=0.9` to give the builder the option to send a JSON response.

Related to https://github.com/ethereum/builder-specs/issues/53 and https://github.com/ethereum/builder-specs/pull/54


metachris:
In support of this change. It's nicely backwards-compatible, doesn't add significant risk, decreases latency and could simplify CL flows.

barnabasbusa:
Would it make sense to mark the json api deprecated with this same PR? That way we can push to get json removed in the next fork.

nflaig:
> That way we can push to get json removed in the next fork.

In my opinion, having JSON as an option is a good thing, it can be quite useful for debugging.

If this reduces complexity significantly for some implementation it could be considered, however in Lodestar for example it does not make a difference at all as we have to keep JSON support for beacon-api anyways.

eserilev:
Im in support of this. In Lighthouse its not a problem for us to support both JSON and SSZ

types/http.yaml
```diff
@@ -21,3 +21,49 @@ InternalError:
       example:
         code: 500
         message: "Internal server error"
+NotAcceptable:
```
nflaig, line 24: Could reuse these error types from beacon-api as https://github.com/ethereum/beacon-APIs/pull/457 has been merged

nflaig:
Lodestar supports this in the latest release ([v1.23.0](https://github.com/ChainSafe/lodestar/releases/tag/v1.23.0)) in case anyone needs a client to test with.

types/http.yaml
ralexstokes: @nflaig do you want to go ahead and make these updates? otherwise I think this is ready to merge

could also do a follow up PR to address this `types/http.yaml` cleanup

apis/builder/header.yaml
```diff
@@ -74,5 +81,7 @@ get:
               value:
                 code: 400
                 message: "Unknown hash: missing parent hash"
+    "406":
+      $ref: "../../builder-oapi.yaml#/components/responses/NotAcceptable"
```
ralexstokes, line 85: do we need 415 here as well?

ralexstokes (commented):
generally looks good!

had one question about adding a 415 code to the headers endpoint

types/http.yaml
nflaig: Having them explicitly here adds a bit more details to the PR if people are just looking at it, on the rendered explorer it will note make a difference.

So for the purpose of this PR might be good to keep as is

apis/builder/header.yaml
nflaig: 415 is returned if the supplied content type [as indicated by the HTTP header](https://github.com/ethereum/builder-specs/pull/104/files#diff-c965c7b0724c38ab5df6102d148d53ab275bac29b52ada697d417c93d2a492fdR56) is not supported by the server, since this endpoint does not accept a request body it's not relevant here

PR #105, opened 2024-10-08
Define electra types and apis 

ensi321:
Add electra type definition and update apis to include new types.

Also update beacon apis submodule commit to `02d595daf39943430f54aa978f88a5c50d957e64`

Follow up of #101


apis/builder/blinded_blocks.yaml
```diff
@@ -50,7 +53,7 @@ post:
             properties:
               version:
                 type: string
-                enum: [ bellatrix, capella, deneb ]
+                enum: [ bellatrix, capella, deneb, electra ]
                 example: "bellatrix"
               data:
                 oneOf:
```
ralexstokes, line 59: do we want to extend the `examples` below here w/ the electra one?

ralexstokes (commented):
had one question but otherwise this looks great, thank you!!

if we update the `beacon-apis` submodule to at least `492bbbdbcad45249d9a8cd8c23e902c5dfe4ad28` then I think this will be final for pectra broadly

ralexstokes:
also looks like the submodule update would help the linter errors

apis/builder/header.yaml
```diff
@@ -44,20 +44,21 @@ get:
...
                  - $ref: "../../builder-oapi.yaml#/components/schemas/Capella.SignedBuilderBid"
                  - $ref: "../../builder-oapi.yaml#/components/schemas/Deneb.SignedBuilderBid"
+                 - $ref: "../../builder-oapi.yaml#/components/schemas/Electra.SignedBuilderBid"
           examples:
             bellatrix:
               $ref: "../../builder-oapi.yaml#/components/examples/Bellatrix.SignedBuilderBid"
             capella:
               $ref: "../../builder-oapi.yaml#/components/examples/Capella.SignedBuilderBid"
             deneb:
-              $ref: "../../builder-oapi.yaml#/components/examples/Deneb.SignedBuilderBid"
```
nflaig, line 60: this looks wrong, shouldn't there be a new example for `electra`

ralexstokes (approved):
lgtm, thanks!

PR #110, opened 2025-01-16
Add SSZ support to register validator api

nflaig:
Follow up on https://github.com/ethereum/builder-specs/pull/104 to also add SSZ support to `registerValidator` api, which should help in cases where the validator client runs a lot of keys and reduce amount of failed requests.

Since we can't really know if the server supports sending the validator registration as SSZ encoded payload the client will be responsible to retry the request using JSON which ideally should only be done if the error status code is 415, indicating an unsupported media type, but since right now we don't really know if all servers properly follow the spec, it is recommended to retry all HTTP errors and after the first error, the client should stop sending SSZ encoded validator registrations.


tbenr (approved):
LGTM

apis/builder/validators.yaml
```diff
@@ -22,6 +22,9 @@ post:
           type: array
           items:
               $ref: "../../builder-oapi.yaml#/components/schemas/SignedValidatorRegistration"
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `List[SignedValidatorRegistrationV1]` bytes. Use content type header to indicate that SSZ data is contained in the request body."
```
jtraglia, line 27:
```suggestion
          description: "SSZ serialized `List[SignedValidatorRegistrationV1, VALIDATOR_REGISTRY_LIMIT]` bytes. Use content type header to indicate that SSZ data is contained in the request body."
```

The SSZ list must have a defined maximum length. I suggest using [`VALIDATOR_REGISTRY_LIMIT`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#state-list-lengths) (essentially unbounded) from the consensus-specs.

nflaig: Good suggestion, I wasn't sure what limit to use so I left it up to implementation but you're right, we should have this in the spec.

PR #113, opened 2025-02-14
Update openapi version and swagger UI

nflaig:
This updates the openapi version and swagger UI to be on par with the beacon-api spec.

The reason why this is required is because the current spec is [invalid](https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fethereum.github.io%2Fbuilder-specs%2Freleases%2Fv0.5.0%2Fbuilder-oapi.json) but applying the changes in this PR [solves this](https://validator.swagger.io/validator/debug?url=https://raw.githubusercontent.com/nflaig/builder-api-spec/refs/heads/main/builder-oapi.json). The issue is that while `oneOf` usage was fixed in https://github.com/ethereum/builder-specs/pull/100, the 3.0.3 openapi spec still handles it as a `oneOf` instead of `anyOf`.

This PR includes the following changes
- update openapi spec from 3.0.3 to 3.1.0
- minor schema changes related to that
- update swagger UI to match [beacon-api/dist](https://github.com/ethereum/beacon-APIs/tree/master/dist)


apis/builder/blinded_blocks.yaml
```diff
@@ -51,7 +51,7 @@ post:
       description: Success response.
       headers:
         Eth-Consensus-Version:
-          $ref: "../../beacon-apis/beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version"
+          $ref: "../../builder-oapi.yaml#/components/headers/Eth-Consensus-Version"
```
nflaig, line 54: this change is kinda unrelated but noticed it while running spec tests, the builder header is narrowed down a bit more as it does not include `phase0` and `bellatrix` in the enum

ralexstokes (approved):
lgtm

PR #117, opened 2025-03-21
builder-specs: Add EIP-7594 (PeerDAS) related changes

0x00101010:


examples/fulu/execution_payload_and_blobs_bundle.json
```diff
@@ -0,0 +1,174 @@
...
+        "commitments": [
+          "0x8dab030c51e16e84be9caab84ee3d0b8bbec1db4a0e4de76439da8424d9b957370a10a78851f97e4b54d2ce1ab0d686f"
+        ],
+        "cell_proofs": [
+          "0xcellproof1",
+          "0xcellproof2",
+          "0xcellproof3",
+          "0xcellproof4",
+          "0xcellproof5",
+          "0xcellproof6",
```
jtraglia, line 44: These should probably be semi-legit values. Maybe the values from this spec test?

https://media.githubusercontent.com/media/ethereum/consensus-spec-tests/refs/heads/master/tests/general/fulu/kzg/compute_cells_and_kzg_proofs/kzg-mainnet/compute_cells_and_kzg_proofs_case_valid_4aedd1a2a3933c3e/data.yaml

0x00101010: Awesome, updated

types/fulu/blobs_bundle.yaml
```diff
@@ -0,0 +1,33 @@
...
+          $ref: "../../beacon-apis/types/primitive.yaml#/Blob"
+        minItems: 0
+        maxItems: 4096
+      commitments:
+        type: array
+        items:
+          $ref: "../../beacon-apis/types/primitive.yaml#/KZGCommitment"
+        minItems: 0
+        maxItems: 4096
+      proofs:
```
g11tech, line 21: we should do a new verison as proofs will not be available and has not been marked optional

also the response from builderapi could be ssz seralized for ssz wireformat support and hence should be ssz container friendly

types/fulu/blobs_bundle.yaml
```diff
@@ -0,0 +1,33 @@
+Fulu:
+  BlobsBundle:
+    type: object
+    description: "The `BlobsBundle` object from the CL Fulu spec"
+    required: [blobs, commitments]
+    # proofs should be present before Fulu
+    # cellProofs should be present after Fulu
```
nflaig, line 7: the api is fork-aware, there should not be optional fields as part of a container, meaning `proofs` should be part of electra properties (or rather deneb since it didn't change) and `cellProofs` should be part of fulu properties

types/fulu/blobs_bundle.yaml
```diff
@@ -0,0 +1,33 @@
...
+          $ref: "../../beacon-apis/types/primitive.yaml#/KZGProof"
+        minItems: 0
+        maxItems: 4096
+      cellProofs:
+        type: array
+        items:
+          $ref: "../../beacon-apis/types/primitive.yaml#/KZGProof"
+        minItems: 0
+        # 128 (cells per blob) * 4096 (blobs)
+        maxItems: 524288
```
jimmygchen, line 33: With the introduction of `MAX_CELLS_PER_EXT_BLOB` in https://github.com/ethereum/consensus-specs/pull/4187

This would become `MAX_CELLS_PER_EXT_BLOB * `MAX_BLOB_COMMITMENTS_PER_BLOCK``

types/fulu/blobs_bundle.yaml
```diff
@@ -22,12 +20,6 @@ Fulu:
...
         minItems: 0
-        maxItems: 4096
-      cellProofs:
-        type: array
-        items:
-          $ref: "../../beacon-apis/types/primitive.yaml#/KZGProof"
-        minItems: 0
-        # 128 (cells per blob) * 4096 (blobs)
-        maxItems: 524288
+        maxItems: 4194304
```
jtraglia: In [the consensus-specs PR](https://github.com/ethereum/consensus-specs/pull/4187), after some feedback, we decided to go with `FIELD_ELEMENTS_PER_EXT_BLOB * MAX_BLOB_COMMITMENTS_PER_BLOCK`. This is `8192 * 4096 =33,554,432`.

jtraglia (approved):
LGTM, thanks @0x00101010!

specs/fulu/builder.md
```diff
@@ -0,0 +1,41 @@
...
+### Extended containers
+
+#### `BlobsBundle`
+
+```python
+class BlobsBundle(Container):
+    commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    # [Modified in Fulu:EIP7594]
+    proofs: List[KZGProof, FIELD_ELEMENTS_PER_EXT_BLOB * MAX_BLOB_COMMITMENTS_PER_BLOCK]
+    blobs: List[Blob, MAX_BLOB_COMMITMENTS_PER_BLOCK]
```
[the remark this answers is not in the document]
0x00101010: Updated

types/fulu/blobs_bundle.yaml
```diff
@@ -0,0 +1,25 @@
...
+        type: array
+        items:
+          $ref: "../../beacon-apis/types/primitive.yaml#/KZGCommitment"
+        minItems: 0
+        maxItems: 4096
+      proofs:
+        type: array
+        items:
+          $ref: "../../beacon-apis/types/primitive.yaml#/KZGProof"
+        # 8192 (FIELD_ELEMENTS_PER_EXT_BLOB) * 4096 (blobs)
```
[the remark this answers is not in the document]
0x00101010: updated

specs/fulu/builder.md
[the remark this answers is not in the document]
jtraglia: We put the comment on the line above when the comment would exceed our line length limit. In this instance it would. But it doesn't really matter here.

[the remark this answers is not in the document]
nflaig: ah got it, I let you handle it @jtraglia, I confused myself as I thought the comment meant that both `proofs` and `blobs` was modified but figured we are only updating proofs limit

jimmygchen (commented):
I'm wondering if it still makes sense for the `submit_blinded_block` endpoint to return the full payload and blobs bundle, especially with the growing response size due to more blobs and proofs.

Even today on mainnet, we regularly see relays not return the revealed response – the relay still publishes the block and blobs, just doesn’t return them in the response. So the response isn’t that useful for the beacon node in practice, particularly if it comes back late or not at all.

With blob counts increasing, this will likely get worse – larger responses make it more likely relays won’t return anything, and it's a lot of overhead to serialise/deserialise something that doesn't add much value.

Maybe it’s time to update this and remove the response body altogether for Fulu – and instead explicitly require the builder to publish the block and blobs? That feels more aligned with actual behaviour, and avoids wasting resources on a response that's often late and unnecessary.

tersec:
> Maybe it’s time to update this and remove the response body altogether for Fulu – and instead explicitly require the builder to publish the block and blobs? That feels more aligned with actual behaviour, and avoids wasting resources on a response that's often late and unnecessary.

This is reasonable, but the spec should also then explicitly require (spec verb: **`MUST`**) the relay/builder system to send out the resulting unblinded block.

alextes:
my 2ct
1. we always return the response. ofc, with the rule that we only return the response once we're sure there's no way the proposer can actually affect much and equivocate. they're still a bit of a p2p broadcast / propagation boost although speaking for us at least we're already set up for fast propagation through [Fiber network](https://fiber.chainbound.io/). plus, given header request delay + signed header return + block publish and propagate, it's virtually always latency in returning the signed header to us that results in missed slot, its exceedingly rare a header request is delayed too much at the outset, or that a signed header is returned in time but our publish or propagation took so long the slot was missed.
2. i'm not currently aware of any reason to keep the full response besides the small propagation boost. as long as the rule is to only return if our own publish is successful, that is naturally what we optimize for.

jimmygchen:
> Maybe it’s time to update this and remove the response body altogether for Fulu – and instead explicitly require the builder to publish the block and blobs? 

Cross posting here for visibility - given we now have EL `getBlobs` which relies on receiving a block / data column sidecar, it might still be worth returning the revealed payload, so the proposer can help propagate the block too.

tersec:
> > Maybe it’s time to update this and remove the response body altogether for Fulu – and instead explicitly require the builder to publish the block and blobs?
> 
> Cross posting here for visibility - given we now have EL `getBlobs` which relies on receiving a block / data column sidecar, it might still be worth returning the revealed payload, so the proposer can help propagate the block too.

I mean, I think that might be a little farcical given:
> we always return the response. ofc, with the rule that we only return the response once we're sure there's no way the proposer can actually affect much and equivocate.

Other relay/builder systems behave similarly.

examples/fulu/execution_payload_and_blobs_bundle.json
```diff
@@ -0,0 +1,173 @@
+{
+  "value": {
+    "version": "deneb",
```
terencechain, line 3: I think the version should be fulu given the blobs bundle v2 changes

ralexstokes:
Going to move the issue around changing the `getPayload` semantics here: https://github.com/ethereum/builder-specs/issues/119

Otherwise, we can go ahead and merge this in soon to have something to work towards as we begin Fusaka development.

PR #128, opened 2025-08-18
Add SSZ examples

nflaig:
- add ssz examples in addition to json
- add `SignedValidatorRegistrationV1` example
- fix lint warnings due to example/schema mismatches
- update examples that contained invalid data

**Note:** I did not reference the ssz examples in the schema itself as swagger doesn't display them correctly, which is a [known issue](https://github.com/ethereum/builder-specs/pull/104#issuecomment-2302039603).


bharath-123:
thanks for this! it would be really nice to have a test script which validates the ssz. Should be easy since we know the types?

nflaig:
>  it would be really nice to have a test script which validates the ssz. Should be easy since we know the types?

It would require to bring in a full ssz library into this repo as the spec right now is not executable. Maybe a script to generate ssz examples from json could be useful but there aren't that many examples so I just manually did it using https://simpleserialize.com/.

bharath-123 (approved):

PR #138, opened 2025-11-28
Staked Builder API for Glamsterdam

### LINKED ISSUE #137: Specify Builder-API with ePBS
With ePBS being SFI'd for Glamsterdam, we will likely see some major updates to the builder-api. 

TLDR; ePBS introduces a mechanism to stake a builder in-protocol and have a mechanism to trustlessly pay a proposer for including an execution payload.
Certain relays/builders might not choose to become a staked builder and might want to still run off-protocol like how it is today. In such scenarios, it would be good for us to understand how the builder-api would change and design it accordingly so that there is a standard for implementing it.

bharath-123:
This PR begins work and discussions on the Builder-API for ePBS for staked builders.

ePBS introduces the following protocol upgrades:
1. Enshrining of builders into the protocol by allowing them to stake collateral (used for trustless payments).
2. Decoupling of `BeaconBlock` validation and `ExecutionPayload` validation.

We introduce 3 new APIs. We ideally want to have new APIs to make the transition from Fulu to Gloas easier:

1. `getExecutionPayloadBid`: POST `/eth/v1/builder/execution_payload_bid/{slot}/{parent_hash}/{parent_root}/{proposer_pubkey}`

   Called by the proposer to the builder/relay. The builder/relay returns the best execution payload bid available to it. Per-request validator inputs are communicated as follows:

   - `SignedRequestAuthV1` request body (**optional**): authenticates the request. The body MAY be encoded as JSON (`Content-Type: application/json`) or SSZ (`Content-Type: application/octet-stream`); when SSZ is used, the `Eth-Consensus-Version` header MUST also be set. If the body is omitted, the builder MAY still serve a bid.

2. `submitBuilderPreferences`: POST `/eth/v1/builder/builder_preferences/{validator_pubkey}`

   Called by the proposer in the epoch prior to the epoch in which they will be proposing. The proposer sends a `BuilderPreferencesRequestV1` containing a `BuilderPreferencesV1` (with `max_execution_payment`) and a `SignedRequestAuthV1` to authenticate the request. The builder MUST check that `auth.message.data` matches its own URL and MUST verify the BLS signature against the `validator_pubkey` path parameter to ensure that the preference came from the intended proposer.

3. `submitSignedBeaconBlock`: POST `/eth/v1/builder/beacon_block`

   Called by the proposer when it commits to the bid received from `getExecutionPayloadBid`. The proposer sends the `SignedBeaconBlock` with the execution payload bid embedded. The builder is then responsible for constructing and broadcasting the corresponding `SignedExecutionPayloadEnvelope` to the PTC committee.

`ValidatorRegistrationV1` is **deprecated** in favour of `ProposerPreferences` from the consensus specs. General validator preferences (`fee_recipient`, `gas_limit`, etc.) are now communicated via the `proposer_preferences` gossip topic defined in the consensus specs. Validators broadcast `SignedProposerPreferences` messages at the beginning of each epoch for their proposal slots in the next epoch, and builders SHOULD subscribe to this gossip topic.

We introduce `RequestAuthV1` which contains the following fields:
1. `data`: MUST be set to the URL of the builder the request is intended for.
2. `slot`: The slot for which the request is being sent.

`SignedRequestAuthV1` is a container holding a `RequestAuthV1` message and a BLS signature over it. The goal is to allow the proposer to authenticate to a specific builder for a given slot, preventing DDOS and replay attacks from other parties. Sending the `SignedRequestAuthV1` with `getExecutionPayloadBid` is optional; builders MAY use it for authentication purposes.

We also introduce `BuilderPreferencesV1`, which carries `max_execution_payment` — the maximum execution layer payment (in Gwei) the proposer will accept from this builder. If no preferences have been submitted, the builder MUST treat `max_execution_payment` as `0` and MUST NOT include an execution layer payment in its bid.

Closes: https://github.com/ethereum/builder-specs/issues/137


apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+    If the builder is unable to produce a valid execution payload bid, then
+    the builder MUST return a 204 response. If the request is invalid, then the
+    builder MUST return an error response (400) with a description of the
+    validation failure.
+
+    This API is applicable from Glamsterdam fork onwards.
+  tags:
+    - Builder
+  parameters:
```
bharath-123: We are no longer passing in the `slot`, `parent_hash` and `pubkey` as params in the path. I am unsure if we need to submit that data anymore since the validator already checks if the latest slot and parent_hash match from state while processing the [ExecutionPayloadBid](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-process_execution_payload_bid)

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,89 @@
...
+      schema:
+        type: integer
+        format: int64
+        example: 1710338135000
+    - name: X-Timeout-Ms
+      in: header
+      required: false
+      description: |
+        Optional header containing the proposer's timeout for the request in milliseconds. 
+        Relays should use this header to adjust the amount of time by which they delay getBid
```
bharath-123:
```suggestion
        Builders should use this header to adjust the amount of time by which they delay getBid 
```

specs/gloas/validator.md
```diff
@@ -0,0 +1,72 @@
...
+  `builder_pubkey`.
+* `proposal_epoch`: This is set to `get_current_epoch(state) + 1`.
+
+
+### Validator Registration dissemination
+
+This specification suggests validators re-submit registrations only if they will be proposing in the upcoming epoch(E+1). 
+This is to avoid sending a lot of `ValidatorRegistrations` every epoch. This can potentially help reduce the load 
+Validators are expected to perform this check at every epoch boundary. Validators can send their registrations even though
+they won't be proposing in the upcoming epoch.
```
bharath-123: we don't want to have a very strong enforcement on validators to send registrations only when they are proposing in epoch E+1 but we recommend them to do this.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,133 @@
...
+        format: int64
+        example: 10000
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+            $ref: "../../beacon-apis/types/gloas/bid_request_auth.yaml#/Gloas/SignedBidRequestAuth"
```
bharath-123: We send a `SignedBidRequestAuth` for each builder index registered by the builder. This could become large if a builder has registered multiple builder indices for themselves. But I am unsure as to why they would do that.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,133 @@
...
+            title: GetExecutionPayloadBidsResponse
+            type: object
+            required: [version, data]
+            properties:
+              version:
+                type: string
+                enum: [ gloas ]
+                example: "gloas"
+              data:
+                type: array
```
bharath-123: We get back a bid corresponding to each builder index. The builder may not choose to return bids for all builder indices.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,125 @@
...
+      schema:
+        type: integer
+        format: int64
+        example: 10000
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: "../../types/gloas/bid_request_auth.yaml#/Gloas/SignedBidRequestAuth"
```
bharath-123: We send one SignedBidRequestAuth per request. If a builder has multiple builder indices associated with them, the client can send a request per builder index.

bharath-123:
Note: We need to fix the links and formatting of links to consensus-specs in this PR.

specs/gloas/validator.md
```diff
@@ -0,0 +1,191 @@
...
+
+Validators request an `ExecutionPayloadBid` from the external builder network to
+put it in their `SignedBeaconBlock`. The external builder network broadcasts the
+`SignedExecutionPayloadEnvelope` corresponding to the bid to the PTC committee.
+
+## Containers
+
+### New Containers
+
+#### `BidRequestAuth`
```
bharath-123: We could probably have a better name here?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,113 @@
...
+      description: Hash of execution layer block the proposer will build on.
+      schema:
+        $ref: "../../builder-oapi.yaml#/components/schemas/Root"
+    - name: parent_root
+      in: path
+      required: true
+      description: Root of the beacon block the proposer will build on.
+      schema:
+        $ref: "../../builder-oapi.yaml#/components/schemas/Root"
+    - name: proposer_index
```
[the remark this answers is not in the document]
nerolation: There are ideas around reusing the proposer index, whereas the proposer pubkey is more in the custody of the proposer (rather than easily changeable with a protocol change).
Thus, pubkey might be preferable if it doesn't add any complexity.

specs/gloas/builder.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Containers
+
+### New Containers
+
+#### `BuilderPreferences`
+
+```python
+class BuilderPreferences(Container):
+    execution_payment_accepted: boolean
```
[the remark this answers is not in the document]
nerolation: What arbitrary preferences are we talking about here? My feeling is that we don't necessarily need that. If proposers want builders to perform arbitrary logic, they just shouldn't connect to any other builder.

specs/gloas/builder.md
```diff
@@ -0,0 +1,206 @@
...
+[Gloas consensus specs][gloas-builder-specs].
+
+## Constructing a `SignedExecutionPayloadEnvelope`
+
+If the builder's [`SignedExecutionPayloadBid`][signed-execution-payload-bid] has
+been accepted by the proposer and it has been included in it's
+`SignedBeaconBlock`, then the builder has to construct a
+[`SignedExecutionPayloadEnvelope`][signed-execution-payload-envelope]
+corresponding to the [`SignedExecutionPayloadBid`][signed-execution-payload-bid]
+and it has to broadcast it to the PTC committee via the
```
nerolation: has to broadcast it via the ... (remove to the PTC committee)

specs/gloas/validator.md
nerolation: Maybe ProposerAuth?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,125 @@
...
+  description: |
+    Requests a builder node to produce a valid execution payload bid, which
+    can be integrated into a blinded beacon block and signed. 
+
+    The proposer sends a POST request to the builder with the following information:
+    - The slot for which the block should be proposed.
+    - The hash of the execution layer block the proposer will build on.
+    - The root of the beacon block the proposer will build on.
+    - The index of the proposer.
+    - A signed bid request auth to authenticate the request to a specific builder index.
```
nerolation: I'm wondering should the API support batching multiple builder registration queries with a single auth, or is one request per builder intentional?

nerolation:
> Open questions:
> 
> 1. How is the `fee_recipient` communicated to the builder? Do we keep `ValidatorRegistrations` or do we send the `fee_recipient` along as a header while receiving the bid. Or we can use the `ProposerPreferences` gossip topic as specified in: [Add specs for proposer preferences consensus-specs#4777](https://github.com/ethereum/consensus-specs/pull/4777)
> 2. The API currently assumes all builders are staked. do we have to also support builders who choose to not stake?

1. Using ValidatorRegistrationV2 seems cleanest since it's already being sent pre-epoch. Adding it as an HTTP header feels like a workaround. In case of not obeying the fee_recipient (same for gas limit) there should be a fallback to "local building".

builder-oapi.yaml
```diff
@@ -102,7 +108,17 @@ components:
...
+      $ref: "./types/gloas/bid_request_auth.yaml#/Gloas/BidRequestAuth"
+    Gloas.SignedBidRequestAuth:
+      $ref: "./types/gloas/bid_request_auth.yaml#/Gloas/SignedBidRequestAuth"
+    Gloas.BuilderPreferences:
+      $ref: "./types/gloas/registration.yaml#/Gloas/BuilderPreferences"
+    Gloas.ValidatorRegistrationV2:
+      $ref: "./types/gloas/registration.yaml#/Gloas/ValidatorRegistrationV2"
+    Gloas.SignedValidatorRegistrationV2:
+      $ref: "./types/gloas/registration.yaml#/Gloas/SignedValidatorRegistrationV2"
+
```
bharath-123: We should ideally move these to the beacon-api repo when things have been agreed upon w.r.t the Builder-API.

specs/gloas/builder.md
```diff
@@ -0,0 +1,196 @@
...
+[`SignedExecutionPayloadEnvelope`][signed-execution-payload-envelope] objects.
+
+## Containers
+
+### New Containers
+
+#### `BuilderPreferences`
+
+```python
+class BuilderPreferences(Container):
```
potuz: I am uncertain about the min-bids, requested inputs from pools. What I think should be included instead of `accept_trusted_payments` is to replace the boolean with a single `max_trusted_bid` indicating what is the maximum amount the validator would take on a trusted bid, any value exceeding this should be trustless. This includes the boolean since setting this value to zero is equivalent to being false on the boolean and setting this value to max uint is equivalent to True; while this still allows the validator to accept risking not receiving funds up to a certain threshold. I can't think of any reason why a validator would not want to make this public and instead hide this value from the builder.

bharath-123: I think this makes sense to me and I am not very opinionated on one way or the other. I think making it uint can add more expressivity.

specs/gloas/validator.md
```diff
@@ -0,0 +1,246 @@
...
+    builder = state.builders[bid.builder_index]
+
+    assert is_active_builder(state, builder)
+    assert bid.slot == state.slot
+    assert bid.fee_recipient == fee_recipient
+    assert bid.parent_block_hash == state.latest_block_hash
+    assert bid.parent_block_root == hash_tree_root(state.latest_block_header)
+    assert bid.prev_randao == get_randao_mix(state, get_current_epoch(state))
+
+    assert bid.execution_payment <= reg.message.builder_preferences.max_trusted_bid
```
faheelsattar: should also validate the `bid.gaslimit` against the `reg.message.gas_limit`

bharath-123:
Point to discuss: Do we want to include builder indices associated with the builder in the Validator registrations and while authenticating the bid request? What makes this tricky to do is that, builder indices associated with a builder could dynamically change and the client would require an API to periodically fetch the latest builder indices and cache it. This generally seems like a pain, so we proposer to not include builder indices in validator registrations and while authenticating the bid request. For the case of authenticating the bid request, we propose signing over the builder url.

apis/builder/execution_payload_bid.yaml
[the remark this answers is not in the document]
eserilev: I believe we use pubkey in the existing `getPayloadHeader` so maybe for consistency we should stick to pubkey.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,125 @@
...
+      required: false
+      description: |
+        Optional header containing a Unix timestamp in milliseconds representing
+        the point-in-time the request was sent. This header can be used to measure
+        latency.
+      schema:
+        type: integer
+        format: int64
+        example: 1710338135000
+    - name: X-Timeout-Ms
```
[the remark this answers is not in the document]
bharath-123: refer to https://ethpandaops.io/posts/fusaka-more-blobs-less-votes/ for an analysis on the relation between timing games and attestation performance degradation as blob count increases with BPOs.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,126 @@
+post:
+  operationId: "getExecutionPayloadBid"
+  summary: Get an execution payload bid.
+  description: |
+    Requests a builder node to produce a valid execution payload bid, which
+    can be integrated into a blinded beacon block and signed.
```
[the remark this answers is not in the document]
potuz: Couldn' t reply to @eserilev direcly, but what is the benefit of sending pubkey vs index post Gloas?

apis/builder/execution_payload_bid.yaml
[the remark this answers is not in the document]
potuz: This call is in the hot path I don't see any reason to send extra bytes. The slot is signed so index reuse is irrelevant. But this is a soft take. I think getting the index is always cheaper than the pubkey at the point of the call, but I empathize with @eserilev's comment that perhaps the caller code needs to be changed less if we stick to pubkey

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,176 @@
...
+      authenticates the request. The body MAY be encoded as JSON
+      (`Content-Type: application/json`) or SSZ
+      (`Content-Type: application/octet-stream`).
+
+    The `X-Eth-Max-Trusted-Bid` header is required. If it is missing or
+    malformed, the builder MUST NOT serve a bid for the proposer and MUST
+    return a 400 response.
+
+    The `SignedRequestAuth` body is optional. If it is present but malformed
+    or fails signature verification, the builder MUST return a 400 response.
```
potuz: Why this MUST? the builder can ignore it if it wants.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,176 @@
...
+        type: integer
+        format: int64
+        example: 1710338135000
+    - name: X-Timeout-Ms
+      in: header
+      required: false
+      description: |
+        Optional header containing the proposer's timeout for the request in milliseconds.
+        Builders should use this header to adjust the amount of time by which they delay the
+        requests to maximise block rewards. Otherwise, requests will timeout and the proposer
```
potuz: why is this even specified?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,176 @@
...
+        payment amount. If the header is missing or malformed, the builder
+        MUST return a 400 response.
+      schema:
+        type: integer
+        format: uint64
+        example: 1000000000
+  requestBody:
+    description: |
+      Optional `SignedRequestAuth` authenticating the request. If provided,
+      the builder MUST verify the BLS signature against `validator_pubkey`
```
potuz: why the MUST? the builder can do whatever they want, this should be a MAY

specs/gloas/builder.md
```diff
@@ -0,0 +1,164 @@
...
+    parent_hash: Hash32,
+    # [New in Gloas]
+    parent_root: Root,
+    # [New in Gloas]
+    validator_index: ValidatorIndex,
+):
+    # Verify slot
+    assert slot == state.slot
+
+    assert validator_index in state.validators.keys()
```
potuz: This would be typically an assert against the length of the slice.

specs/gloas/builder.md
```diff
@@ -0,0 +1,164 @@
...
+[`ProposerPreferences`][proposer-preferences], regardless of whether the builder
+pays via `bid.execution_payment` or `bid.value`.
+
+If the builder intends to pay the proposer via their staked collateral, they
+MUST set `bid.value` to the amount they are committing to pay.
+
+If the builder intends to pay the proposer via an execution layer payment, they
+MUST set `bid.execution_payment`. This value MUST NOT exceed the
+`max_trusted_bid` received in the `X-Eth-Max-Trusted-Bid` header of the
+corresponding [`getExecutionPayloadBid`][get-execution-payload-bid-api] request.
```
potuz: Perhaps a not in this section that any value in `bid.value`  will be deducted irrespectively of having or not any value in `bid.execution_payment`, to prevent the confusion that some builder may think these fields are exclusive.

specs/gloas/validator.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### `RequestAuth`
+
+`RequestAuth` is used to authenticate requests to a builder. This is useful so
+that other builders do not DDOS or run replay attacks on the builder.
+
+```python
+class RequestAuth(Container):
+    builder_pubkey: BLSPubkey
+    validator_pubkey: BLSPubkey
```
potuz: Why do we need this field? the container that carries the auth already has the proposer index.

specs/gloas/validator.md
```diff
@@ -0,0 +1,217 @@
...
+builder MAY still serve a bid.
+
+### Constructing the `RequestAuth`
+
+If the validator chooses to authenticate its request, it constructs a
+`RequestAuth` with the following fields:
+
+- `builder_pubkey`: The BLS public key of the builder the request is intended
+  for.
+- `validator_pubkey`: The BLS public key of the validator making the request.
```
potuz: Same as above, I think this field can be omitted.

specs/gloas/validator.md
```diff
@@ -0,0 +1,217 @@
...
+the builder will not serve a bid for the proposer.
+
+The validator MAY additionally send a [`SignedRequestAuth`](#signedrequestauth)
+as the request body to authenticate the request. The body MAY be encoded as JSON
+(`Content-Type: application/json`) or SSZ
+(`Content-Type: application/octet-stream`); when SSZ is used, the validator MUST
+also send the `Eth-Consensus-Version` header. If the body is omitted, the
+builder MAY still serve a bid.
+
+### Constructing the `RequestAuth`
```
potuz: A comment in the requestAuth. As stated in this PR, there are two options: either the signature is there for a specific builder or is not there at all. 

A third option is possible which is to include something like `0x00` for the builder. In which case the builder can get the guarantee that the signature is from the validator, without the signature being binding for that particular builder. This can have its own uses if proposers and builders want to use this.

specs/gloas/validator.md
```diff
@@ -0,0 +1,217 @@
...
+### New Containers
+
+#### `RequestAuth`
+
+`RequestAuth` is used to authenticate requests to a builder. This is useful so
+that other builders do not DDOS or run replay attacks on the builder.
+
+```python
+class RequestAuth(Container):
+    builder_pubkey: BLSPubkey
```
potuz: Why do we use the builder pubkey instead of the builder index? this is in the hot path and cannot be replayed so index reuse is useless here.

apis/builder/execution_payload_bid.yaml
bharath-123: yeah this doesn't have to be specced. I was doubting this when I wrote it down.

apis/builder/execution_payload_bid.yaml
bharath-123: This was something added to MEV-Boost by Obol here: https://github.com/flashbots/mev-boost/pull/817 , it made sense to specify it in the builder-spec.

We specify it today too: https://github.com/ethereum/builder-specs/blob/main/apis/builder/header.yaml#L46 

This is also a completely optional header which clients don't have to send.

apis/builder/execution_payload_bid.yaml
bharath-123: yeah fair, i was thinking that we would want such validation but I am fine not having this condition.

specs/gloas/builder.md
bharath-123: yeah thats a good point to have in the spec.

specs/gloas/validator.md
bharath-123: this is so that the builder can verify the signature against this pubkey. I think you mean the builder can just refer to it from validator index pass in the endpoint query params. I think that makes sense to me. We can remove the validator_pubkey from the `RequestAuth` in this case.

specs/gloas/validator.md
bharath-123: Builders can advertise their builder pubkey before the fork and not have to wait for their deposits to succeed. So a validator can just add the builders pubkey and url to their beacon node configuration. This just seemed more convenient to me than adding the builder index.

specs/gloas/validator.md
bharath-123: yeah this is a good idea. this is an option which node operators can use, if they want. To do this, they can just specify 0x000 pubkey instead of the actual builder pubkey in their config where they specify the builders. Its an option for node operators.

JasonVranek: Signing over generic 0x00... lets builders replay your signature which defeats the purpose of the auths. If the concern is too many sigs in the hot-path, the proposer could always pre-sign, so I don't see a ton of value in explicitly adding this to the spec.

Assuming we don't, there's a nice simplification available. The SignedRequestAuth​ doesn't need to live in the body at all. The request can simply attach an optional header like X-Eth-Request-Auth-Signature: 0xabcd... which avoids the JSON/SSZ content negotiation for the request body entirely. Since the builder already knows its own pubkey and can read slot from the path params, it can reconstruct a RequestAuth​ locally to verify. 

Also related, since this is a new API I'd advocate for SSZ-only content encoding for the new endpoints

potuz: > Signing over generic 0x00... lets builders replay your signature which defeats the purpose of the auths.

Some proposers may prefer this, and some builders were requesting the whole feature to not even exist, so this gives both proposers and builders more options. This is a way of making sure that the request actually came from a proposer, but not for which builder. 

Builders may, if so they desire, only take requests that are signed to them. Proposers, if so they desire, may send signed requests specific to each builder. Notice that the existence of the "max_trusted_bid" which is also personal to each builder, already makes for some nuances regarding bid request replaying.

JasonVranek: > This is a way of making sure that the request actually came from a proposer, but not for which builder.

This is a proof of existence not proof of provenance? I agree it provides optionality but if it's as gameable as no auth and builders deny it then the option feels like bloat

specs/gloas/validator.md
```diff
@@ -0,0 +1,225 @@
...
+in favor of [`ProposerPreferences`][proposer-preferences] from the consensus
+specs.
+
+General validator preferences are now communicated via the
+[`proposer_preferences`][proposer-preferences-topic] gossip topic defined in the
+[Gloas consensus specs][gloas-consensus-specs]. At the beginning of each epoch,
+validators broadcast [`SignedProposerPreferences`][proposer-preferences]
+messages for their proposal slots in the next epoch.
+
+Builders SHOULD subscribe to this gossip topic to learn about proposer
```
tersec, line 163: Should one read this as indicating it's safe for CLs to stop using `ValidatorRegistrationV1` at the Gloas boundary entirely?

This `deprecated`/`SHOULD` language isn't really strong enough. Either they exist, in theory, and CLs will have to do this just in case, or they're simply not part of the Gloas builder API, and it's unclear therefore where else builders would get the requisite information.

`SHOULD` is less problematic than the `deprecated` language in this regard.

specs/gloas/validator.md
```diff
@@ -0,0 +1,225 @@
...
+preferences for upcoming slots.
+
+Per-builder preferences (currently just `max_trusted_bid`) are not gossiped;
```
tersec: The spec should use an "eternal present" here, probably: right now it's (no `just`) `max_trusted_bid`. If that changes, then the spec changes. The spec doesn't have to hedge or guess about its own future developments. It might be cleaner to omit the list of per-builder preferences if it's specified elsewhere (and it should be, as a general point; it should not just be in a random parenthetical).

specs/gloas/validator.md
```diff
@@ -0,0 +1,225 @@
...
+
+    return verify_execution_payload_bid_signature(state, signed_bid)
+```
+
+`max_trusted_bid` is the value the validator sent in the `X-Eth-Max-Trusted-Bid`
+header of the corresponding
+[`getExecutionPayloadBid`][get-execution-payload-bid-api] request. Validators
+MUST validate each bid against the `max_trusted_bid` they sent for that request.
+
+Note that, the fee recipient specified in `bid.fee_recipient` does not
```
tersec:
```suggestion
Note that the fee recipient specified in `bid.fee_recipient` does not
```

specs/gloas/validator.md
tersec: Conversely, are we expecting that in practice builders will still allow `ValidatorRegistrationV1` as a kind of sidechannel, and that's why the spec is written in this way?

specs/gloas/validator.md
bharath-123: This is not something we have to explicitly spec out anyways. If a proposer desires to send the same 0x000 builder pubkey to all builders, they could just explicitly set the pubkey associated with the builder in the configs to 0x000. So this is optionality left to the validator and not something the spec has to be explicit about. 
RequestAuth is more for builder privacy and safety rather than for proposers.

specs/gloas/validator.md
bharath-123: I can update the language here. 

The spec requires validators to broadcast their preferences an epoch before the fork: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md?plain=1#L431 so at the gloas fork boundary I don't think there is any requirement to send `ValidatorRegistrationV1` to the builders, since the proposer preferences contains the same information. 

I don't see any value for builders to have `ValidatorRegistrationV1` as a sidechannel, its just the same information as that of proposer preferences.

specs/gloas/validator.md
bharath-123: I think this line itself is unnecessary to specify in general now that I think about it. 

Right now the only per builder preferences that makes sense is the max_trusted_bid.

specs/gloas/validator.md
bharath-123: I have explicitly specified it which I don't think is necessary.

apis/builder/beacon_block.yaml
```diff
@@ -0,0 +1,50 @@
+post:
+  operationId: "submitSignedBeaconBlock"
+  summary: Submit a signed beacon block with the execution payload bid.
+  description: |
+    Submits a `SignedBeaconBlock` to the builder, binding the proposer to the block. 
+
+    A success response (200) indicates that the signed beacon block was
```
0w3n-d: Below in responses, it says 202

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+    - The root of the beacon block the proposer will build on.
+    - The index of the proposer.
+    - The `X-Eth-Max-Trusted-Bid` header carrying the proposer's
+      `max_trusted_bid` (decimal `uint64`, in Gwei) for this request.
+    - Optionally, a `SignedRequestAuth` in the request body that
+      authenticates the request. The body MAY be encoded as JSON
+      (`Content-Type: application/json`) or SSZ
+      (`Content-Type: application/octet-stream`).
+
+    The `X-Eth-Max-Trusted-Bid` header is required. If it is missing or
```
0w3n-d: Why is this a header rather than just another field in the body?

examples/gloas/signed_request_auth.json
```diff
@@ -0,0 +1,9 @@
+{
+  "value": {
```
0w3n-d, line 2: Why is this "value" needed? Shouldn't is just be message and signature?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+            $ref: "../../builder-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            InvalidHash:
+              value:
+                code: 400
+                message: "Unknown hash: missing parent hash"
+            InvalidAuth:
+              value:
+                code: 400
+                message: "Invalid SignedRequestAuth: signature verification failed"
```
0w3n-d: This should probably be 401 as well no?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+
+    If the builder is unable to produce a valid execution payload bid, then
+    the builder MUST return a 204 response. If the request is invalid, then the
+    builder MUST return an error response (400) with a description of the
+    validation failure.
+
+    This API is applicable from Gloas fork onwards.
+  tags:
+    - Builder
+  parameters:
```
0w3n-d, line 38: In the existing api the query params make sense as it's a GET, but this new one is a POST with a mix of query params, a body, and headers that contain application data rather than just meta data.

Shouldn't it all just be in the body if it's a POST?

apis/builder/beacon_block.yaml
```diff
@@ -0,0 +1,50 @@
...
+  requestBody:
+    description: A `SignedBeaconBlock`.
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          required: [data]
+          properties:
+            data:
```
nflaig: we usually don't use a `data` container in the request body

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+        Builders should use this header to adjust the amount of time by which they delay the
+        requests to maximise block rewards. Otherwise, requests will timeout and the proposer
+        will not receive the header in time.
+      schema:
+        type: integer
+        format: int64
+        example: 10000
+    - name: Eth-Consensus-Version
+      in: header
+      required: false
```
nflaig: why is this header not required? if we change `SignedRequestAuth` in the future it has to be present

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+      description: |
+        The active consensus version to which the `SignedRequestAuth` in the
+        request body belongs. Required if the request body is SSZ encoded.
+      schema:
+        $ref: "../../builder-oapi.yaml#/components/schemas/ConsensusVersion"
+    - name: X-Eth-Max-Trusted-Bid
```
nflaig: should this rather be a query param, usually we only sent metadata in headers, could also consider making this optional, and omitting it means only trustless bids are accepted, same as setting to 0

specs/gloas/builder.md
```diff
@@ -0,0 +1,171 @@
...
+## Bidding
+
+In Gloas, Execution payloads are built for a specific `slot`, `parent_hash`,
+`validator_index` along with the `parent_root` tuple corresponding to a unique
+beacon block serving as the parent.
+
+This is because in Gloas with [EIP-7732], the execution payload and beacon
+blocks are decoupled. The `parent_hash` could refer to a beacon block which is
+an ancestor of the parent beacon block corresponding to the current beacon block
+for which we are building the execution payload.
```
0w3n-d: I think ` corresponding to the current beacon block for which we are building the execution payload` is not needed here. It's implied and easier to read when removed.

specs/gloas/builder.md
```diff
@@ -0,0 +1,171 @@
...
+signature against the validator pubkey resolved from the `proposer_index` path
+parameter, and check that `builder_pubkey` matches their own identity and that
+`slot` matches the requested slot. If verification fails, the builder MAY return
+a 400 response.
+
+If the request body is absent, the builder MAY still serve a bid.
+
+### `max_trusted_bid`
+
+`max_trusted_bid` is the maximum value (in Gwei) that a proposer is willing to
```
0w3n-d: Is it possible to move this to the proposer_preferences?

As it stands the builder/relay only learns of what the max bid is at the moment the proposer calls for the bid.
This means builders either need to wait until the request arrives to commence building the block, build many blocks covering a range of possible max bid values and then respond with the highest bid that is less then the max bid, or use a side channel to obtain the proposers `max_trusted_bid` ahead of time.

Builders/relays with a better relationship with node operators would have an advantage by being in a position to acquire this vital piece of information ahead of time, out of band.

specs/gloas/validator.md
```diff
@@ -0,0 +1,214 @@
...
+    signature: BLSSignature
+```
+
+## Bid Request
+
+When calling [`getExecutionPayloadBid`][get-execution-payload-bid-api], the
+validator MUST send the `X-Eth-Max-Trusted-Bid` header carrying a decimal
+`uint64` (in Gwei) expressing the per-builder `max_trusted_bid` for this
+request. See [`max_trusted_bid`](#max_trusted_bid). If the header is missing,
+the builder will not serve a bid for the proposer.
```
0w3n-d: If the header is missing couldn't the builder still use the `bid.value` instead?
So it's absence just falls back to a default value of zero and but the trustless path still remains intact?

Not suggesting it should be optional but maybe it would be better fall back to trustless rather than nothing at all.

apis/builder/beacon_block.yaml
bharath-123: I ll remove that!

specs/gloas/validator.md
bharath-123: yeah that makes sense to me, we can just fallback to trustless. I assumed builders would prefer to ensure that they pay via the trusted path as much as they can. But we can just fallback to trustless here.

specs/gloas/builder.md
bharath-123: I was thinking about adding this to the proposer preferences. My initial thoughts were that this should be a value which is per-builder. But I believe that can also get operationally challenging for an operator to decide what value of `max_trusted_bid` to set for each builder. 
If it is not necessary that this is per-builder, i can see a good case to move this to the proposer preferences.

apis/builder/execution_payload_bid.yaml
bharath-123: 401 is if the authorization fails or is not valid. I think this makes sense to be a 401.

examples/gloas/signed_request_auth.json
bharath-123: That seems to be the common pattern in writing examples in the builder-specs. To encapsulate them in `value`. I am not too opinionated on this honestly, we can revisit this nit later.

apis/builder/beacon_block.yaml
bharath-123: Yeah my bad, i ll fix this

specs/gloas/builder.md
JasonVranek: The concern about per-request `max_trusted_bid` is that builders may need to readjust their trusted/untrusted payment split on the fly. But in practice, this adjustment is swapping one tx and resealing the block, and builders will have on the order of seconds to do it (simpler than negotiating side channels). 

More importantly, the builders who’d be affected by this are the ones trying to optimize their on-chain staked eth to liquid eth ratio (i.e., they could just default to sending trustless bids all the time and be unaffected). Instead of changing the specs to cater towards this category of builder, I feel it’s better to cater to proposers will have different risk apetites per builder. Forcing a blanket `max_trusted_bid` removes this optionality from the proposer.

note I am unopinionated as to how the max_trusted_bid​ is communicated (header / param / body) but feel it should be set per-request

apis/builder/execution_payload_bid.yaml
bharath-123: Yeah I am unopinianated on this. We can discuss this again.

apis/builder/execution_payload_bid.yaml
bharath-123: I am not too opinianated on this. We can revisit these aspects.

apis/builder/execution_payload_bid.yaml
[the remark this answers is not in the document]
AntiD2ta: In favour of simplifying client implementation by sticking to pubkeys

types/gloas/request_auth.yaml
```diff
@@ -0,0 +1,19 @@
+Gloas:
+  RequestAuth:
+    type: object
+    required: [builder_pubkey, slot]
+    properties:
+      builder_pubkey:
```
AntiD2ta: How should the proposer learn about the builder's pubkey? Or verify it

Could `/eth/v1/builder/status` be extended, or a new endpoint be added for this purpose?

specs/gloas/validator.md
```diff
@@ -0,0 +1,271 @@
...
+
+*Note*: Validator registrations (`ValidatorRegistrationV1`) are **deprecated**
+in favor of [`ProposerPreferences`][proposer-preferences] from the consensus
+specs.
+
+General validator preferences are now communicated via the
+[`proposer_preferences`][proposer-preferences-topic] gossip topic defined in the
+[Gloas consensus specs][gloas-consensus-specs]. At the beginning of each epoch,
+validators broadcast [`SignedProposerPreferences`][proposer-preferences]
+messages for their proposal slots in the next epoch.
```
AntiD2ta, line 161: I'm worried this wording may suggest that broadcasting the proposer's preferences (through the gossip topic) is a validator/proposer responsibility rather than a beacon node responsibility. Validators must make sure to publish their preferences to a beacon node for it to broadcast them, althought it is not clear how.

Vouch and similar validator clients do not run libp2p gossip stacks. They rely on beacon nodes to publish on their behalf. This PR deprecates V1 registrations but does not specify which beacon-node endpoint accepts `SignedProposerPreferences` for publishing.

apis/builder/beacon_blocks.yaml
```diff
@@ -0,0 +1,52 @@
...
+          required: [data]
+          properties:
+            data:
+              $ref: "../../beacon-apis/types/gloas/block.yaml#/Gloas/SignedBeaconBlock"
+              description: "The signed beacon block."
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `SignedBeaconBlock` bytes. Use content type header to indicate that SSZ data is contained in the request body."
+  responses:
+    "202":
```
AntiD2ta, line 36: What is the proposer expected to do after 202?  Wait for gossip envelope?  Publish itself?  Assume the block will be published?

Is the recommendation on what the proposer should do best after getting the 202 documented somewhere?

bharath-123: Its the builders responsibility to publish the envelope in ePBS. The proposer has no way to build the envelope since the payload is with the builder.  The proposer has actually nothing to do post this as the publishing of envelope is completely on the builder. 
The proposer can listen to the PTC votes to see if it was revealed in a timely fashion or not.

specs/gloas/validator.md
bharath-123: The proposer preferences are gossiped by the beacon node based on the proposer lookahead. More information is here: https://github.com/ethereum/consensus-specs/blame/master/specs/gloas/validator.md#L126

The beacon node handles publishing the preferences if the validators attached to the node are proposing in the upcoming epoch based on the proposer lookahead. 

I would assume vouch wouldn't have to do much here? as the underlying beacon node would take care of the proposer preferences publishing?

types/gloas/request_auth.yaml
bharath-123: The builder would have to advertise their pubkey along with the url here. 

Extending `/eth/v1/builder/status` is not a bad idea. We could get back the builder pubkey associated with the builder here.

specs/gloas/builder.md
eserilev: The spec says

> The validator MAY submit its [BuilderPreferencesRequest](https://github.com/ethereum/builder-specs/blob/epbs-updates/specs/gloas/validator.md#builderpreferencesrequest) to each builder via the [submitBuilderPreferences](https://github.com/ethereum/builder-specs/blob/epbs-updates/apis/builder/builder_preferences.yaml) API call in the epoch prior to the epoch in which they will be proposing, as determined from state.lookahead. This ensures builders have the preferences before the bid request arrives.

So i dont think theres a timing issue here.

We could include it as a field in proposer preferences if that makes it easier from an implementation perspective. 

I could maybe see an argument for wanting this to be per builder, at least theoretically a proposer might have varying levels of trust for different builders.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+  description: |
+    Requests a builder node to produce a valid execution payload bid, which
+    can be integrated into a beacon block and signed.
+
+    The proposer sends a POST request to the builder with the following information:
+    - The slot for which the block should be proposed.
+    - The hash of the execution layer block the proposer will build on.
+    - The root of the beacon block the proposer will build on.
+    - The index of the proposer.
+    - The `X-Eth-Max-Trusted-Bid` header carrying the proposer's
```
eserilev: I'm not a big fan of the `X-Eth-Max-Trusted-Bid`. If we already have the builder preference flow, why do we need this header? If a validator hasnt submitted builder preference I think we should just assume max trusted bid is u64::max.

apis/builder/builder_preferences.yaml
```diff
@@ -0,0 +1,61 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: "../../types/gloas/builder_preferences.yaml#/Gloas/BuilderPreferencesRequest"
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `BuilderPreferencesRequest` bytes. Use content type header to indicate that SSZ data is contained in the request body."
+  responses:
+    "200":
```
hangleang:
```suggestion
    "202":
```
Inconsistent with `submitSignedBeaconBlock`, also match the status code in the description

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
+post:
```
hangleang, line 1: what is the reason of using `POST` for a get operation? if `POST` request is required, what is the reason for putting those 4 params in the path segments, rather than in request body?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+        A value of `0` indicates that the proposer does not accept any trusted
+        payments from this builder. A value of `2**64 - 1` (`MAX_TRUSTED_BID`)
+        indicates that the proposer accepts any trusted payment amount.
+      schema:
+        type: integer
+        format: uint64
+        example: 1000000000
+  requestBody:
+    description: |
+      Optional `SignedRequestAuth` authenticating the request. If provided,
```
hangleang: sorry for this dumb question, how can a proposer learn about `builder_pubkey` to construct the `RequestAuth`?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,188 @@
...
+      description: Authentication required.
+      content:
+        application/json:
+          schema:
+            $ref: "../../builder-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            MissingAuth:
+              value:
+                code: 401
+                message: "Missing SignedRequestAuth: this builder requires authenticated requests"
```
hangleang: if `SignedRequestAuth` is optional, why there is error for missing it?

specs/gloas/builder.md
```diff
@@ -0,0 +1,196 @@
...
+| `MAX_TRUSTED_BID` | `2**64 - 1` |
+
+## Bidding
+
+In Gloas, Execution payloads are built for a specific `slot`, `parent_hash`,
+`validator_index` along with the `parent_root` tuple corresponding to a unique
+beacon block serving as the parent.
+
+This is because in Gloas with [EIP-7732], the execution payload and beacon
+blocks are decoupled. The `parent_hash` could refer to a beacon block which is
```
hangleang: `parent_hash` is refer to an execution-layer block, not a beacon block. I think this sentence should be rephrased

apis/builder/execution_payload_bid.yaml
bharath-123: Yes I agree with not having the header. If the validator has not submitted the preference, I think we should just assume max trusted bid is 0 and all payments are trustless? But I have a PR to remove the header here: https://github.com/ethereum/builder-specs/pull/161/changes#diff-a186f4bbbe96e38c7699013ced6e58643ff9a03dddbc4fd7584c817920f5ef41L99

apis/builder/execution_payload_bid.yaml
AntiD2ta: My understanding is that it is optional for the proposer to send the request with a `SignedRequestAuth`, but the spec also says that _if it is absent, the builder MAY still serve a bid_.

Shall we also include a 403 error for the case where the signature was provided, but the signature is invalid, and the builder requires auth?

types/gloas/request_auth.yaml
AntiD2ta: I'm afraid I didn't quite understand the mechanics of how the builder can advertise their pubkeys in the URL. Would this come from the protocol? 

A "it is not clear yet" is a valid answer, but in that case, I would push more for supporting sharing the pubkey via the builder API.

AntiD2ta: Do we expect relays to do this advertising work (and perhaps aggregate builders)?

apis/builder/execution_payload_bid.yaml
bharath-123: I used a post cause we are sending the request auth in the body. I put the params in the path segments cause it was what we are doing previously. Let me think more on whether we should put the params in path segments or request body. This point was raised by a builder too

apis/builder/execution_payload_bid.yaml
bharath-123: Not at all a dumb question. A builder should advertise their `builder_pubkey` along with their URL similar to how we know the relay pubkey today along with the url. like:
```
https://0xa15b52576bcbf1072f4a011c0f99f9fb6c66f3e1ff321f11f461d15e31b1cb359caa092c71bbded0bae5b5ea401aab7e@aestus.live
```

apis/builder/execution_payload_bid.yaml
bharath-123: Some builders might require it, in that case they could send an error back instead of sending a bid back. 
The intention of this error code is to offer a standard error code and message for these builders to return if they indeed require the auth.

specs/gloas/builder.md
bharath-123: I am removing this sentence: https://github.com/ethereum/builder-specs/pull/161/changes#diff-221ca7d5a71aa06ecab51f0872030eb8ba2217afd79b1dd24f4987e0282b9004L39 cause its too confusing lmao

AntiD2ta: hmm, I think there is value in that sentence, I find it highly informative. If we keep it, I agree it should be rephrased.

apis/builder/execution_payload_bid.yaml
AntiD2ta: What about an error code to explicitly state that the signature is invalid?

specs/gloas/validator.md
AntiD2ta: Per your answer, vouch wouldn't have much to do here apart from building the preferences and communicating them to the beacon node. 

Still, I don't see either the builder or the gloas-consensus spec define the beacon node endpoint the validator client should use for the beacon client to publish the preferences on the `proposer_preferences` topic.

`register_validator` was the analog for V1, there appears to be no Gloas equivalent in beacon-APIs yet.

apis/builder/execution_payload_bid.yaml
hangleang: what if there is an endpoint say `GET /eth/v1/builder/identity` or any suitable one to return the `builder_pubkey` as well as a field to indicate whether the builder require `SignedRequestAuth` from proposer to request payload bid?

apis/builder/execution_payload_bid.yaml
hangleang: can we just pass the signature in the request header? because builder must know its own pubkey and `slot` is already passed in the path segments, so builder can construct the message, and verify with the signature in the header

specs/gloas/builder.md
bharath-123: I am removing this para in https://github.com/ethereum/builder-specs/pull/161/changes#diff-221ca7d5a71aa06ecab51f0872030eb8ba2217afd79b1dd24f4987e0282b9004L39

specs/gloas/builder.md
bharath-123: I don't think it should be part of the proposer preferences object since this is a per-builder field and that shouldn't be on a global topic.

specs/gloas/validator.md
bharath-123: I think this would be beacon-api you would have to use: https://github.com/ethereum/beacon-APIs/pull/608/changes#diff-faf0bd13f84da9cddead43dc8482896e52d1d19db09a2d688b6fa18f977ec003R1

bharath-123: i have also raised this in discord: https://discord.com/channels/595666850260713488/710831784991916072/1509121450500423691 to confirm whether the beacon-api i linked above is indeed the right one.

types/gloas/request_auth.yaml
bharath-123: yikes hahaha i wanted to paste a URL but I failed to do so hahaha. It should be something similar to how we have ethstaker lists: https://ethstaker.org/mev-relay-list and how we have a relay advertise something like:
```
https://0xa15b52576bcbf1072f4a011c0f99f9fb6c66f3e1ff321f11f461d15e31b1cb359caa092c71bbded0bae5b5ea401aab7e@aestus.live
```
This is how I think the pubkey would be advertised post ePBS by the builder/relay.

bharath-123: Having the builder return the pubkey via the status API is not a bad idea again. I think we can consider doing that.

apis/builder/execution_payload_bid.yaml
bharath-123: that does make sense to have! I will add that.

apis/builder/execution_payload_bid.yaml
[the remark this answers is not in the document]
bharath-123: have changed it to pubkey here: https://github.com/ethereum/builder-specs/pull/161/changes#diff-a186f4bbbe96e38c7699013ced6e58643ff9a03dddbc4fd7584c817920f5ef41R60

apis/builder/execution_payload_bid.yaml
bharath-123: An identity endpoint makes a lot of sense and is better than embedding this info into the status endpoint. I will spec this endpoint.

apis/builder/beacon_blocks.yaml
AntiD2ta: Thanks, makes sense. Could the spec itself capture this? Something along the lines of:

> After receiving 202, the proposer takes no further action on the block; the builder publishes the envelope. The proposer MAY observe PTC votes to monitor timely reveal.

Currently, it makes sense for builder-API clients to race relays for unblinding; therefore, this new clear responsibility switch is a significant control-flow change for any current builder-API client worth highlighting if we prefer being explicit.

specs/gloas/validator.md
AntiD2ta: Thank you, this shed some good light! When the final new beacon node's endpoint gets settled, could we update this description with something along the lines of:

> General validator preferences are now communicated via the
[`proposer_preferences`][proposer-preferences-topic] gossip topic defined in the
[Gloas consensus specs][gloas-consensus-specs]. At the beginning of each epoch,
validators broadcast [`SignedProposerPreferences`][proposer-preferences]
messages _**via <Beacon API's new endpoint>**_ for their proposal slots in the next epoch.

apis/builder/execution_payload_bid.yaml
hangleang: cool! when will we have a last call to get this PR sort out? It's been a while now, I'm still hesitate to implement since there is still stuff moving around

bharath-123: have a spec here: https://github.com/ethereum/builder-specs/pull/161/changes#diff-c48bd203974d43dd649472eb1c21dc906548942662b5f1fb998cf65f2990aaefR1 

I think this might make sense?

specs/gloas/validator.md
AntiD2ta: +1 to tightening the language. From an implementer perspective, a clean cutover at the fork (MUST NOT send V1 in Gloas, equivalently MUST send SignedProposerPreferences only) is easier and safer than running a dual-write period that ages out at some unspecified later point. The current SHOULD leaves ambiguity about whether builders are expected to keep accepting V1 indefinitely as a fallback.

specs/gloas/builder.md
bharath-123: https://github.com/ethereum/builder-specs/pull/161/changes#diff-221ca7d5a71aa06ecab51f0872030eb8ba2217afd79b1dd24f4987e0282b9004R40 this should be better i think? High level is that, the parent hash doesn't uniquely identify a beacon block and we can build on the block hash of a different beacon block in case the proposer wants to reorg the payload.

apis/builder/execution_payload_bid.yaml
bharath-123: We already have an error code for an invalid signature but I made it a 401 here: https://github.com/ethereum/builder-specs/pull/161/changes#diff-a186f4bbbe96e38c7699013ced6e58643ff9a03dddbc4fd7584c817920f5ef41R160

apis/builder/execution_payload_bid.yaml
bharath-123: If in the future, we update the request auth the builder would probably need to try multiple structs to verify the signature. Just sending the full body with the signature seems much easier to me to avoid this issue?

specs/gloas/validator.md
bharath-123: hmmmm I am not sure if we would want to have a beacon api endpoint in the builder-specs? I think this is an implementation detail which doesn't need to be covered by the builder-specs?

AntiD2ta: If this feels like an unnecessary implementation detail, perhaps an alternative could be "**_via the beacon API_**" or similar, just to highlight that it is not a builder-API client's or a validator client's role to handle the gossip mechanics of broadcasting the proposer preferences. Wdty

specs/gloas/builder.md
AntiD2ta: Reads better. Perhaps we should let it rest for a while or wait for other folks to agree before merging this sentence into `main`.

apis/builder/execution_payload_bid.yaml
bharath-123: I am not sure yet about the last call for this PR haha, will raise it in ACDC this week!

bharath-123: yeah I wouldn't recommend to start implementation on this as of now yet.

specs/gloas/builder.md
```diff
@@ -0,0 +1,212 @@
...
+def is_eligible_for_bid(
+    state: BeaconState,
+    proposer_preferences: Dict[ValidatorIndex, ProposerPreferences],
+    slot: Slot,
+    parent_hash: Hash32,
+    # [New in Gloas]
+    parent_root: Root,
+    # [New in Gloas]
+    proposer_pubkey: BLSPubkey,
+):
```
jtraglia, line 61: Not all parameter changes here are documented. Like `proposer_preferences` is new too.

https://github.com/ethereum/builder-specs/blob/1bf8f4910e904e92144f067c6807072f4e09c90d/specs/bellatrix/builder.md?plain=1#L279-L285

Should document what's removed as well. In consensus-specs, we'd do:

```python
# [Modified in Gloas:EIP7732]
# Removed `registrations`
# [New in Gloas:EIP7732]
proposer_preferences: Dict[ValidatorIndex, ProposerPreferences],
...
```

specs/gloas/builder.md
```diff
@@ -0,0 +1,212 @@
...
+    validator_index = get_beacon_proposer_index(state)
+    assert state.validators[validator_index].pubkey == proposer_pubkey
+
+    # Verify the proposer is active
+    assert is_active_validator(
+        state.validators[validator_index], get_current_epoch(state)
+    )
+
+    # Verify that proposer preferences have been received via the gossip topic
+    assert validator_index in proposer_preferences.keys()
```
jtraglia, line 76: This is modified but not marked as such.

specs/gloas/builder.md
```diff
@@ -0,0 +1,212 @@
...
+    # their view of the chain.
+    # [Modified in Gloas:EIP7732]
+    assert (
+        parent_hash == state.latest_execution_payload_bid.block_hash
+        or parent_hash == state.latest_block_hash
+    )
+
+    # Verify parent root
+    # [Modified in Gloas:EIP7732]
+    assert parent_root == hash_tree_root(state.latest_block_header)
```
jtraglia, line 88: This is completely new, not modified.

specs/gloas/validator.md
```diff
@@ -0,0 +1,258 @@
...
+    auth: SignedRequestAuthV1
+```
+
+## Submitting Builder Preferences
+
+The validator MAY submit its
+[`BuilderPreferencesRequestV1`](#builderpreferencesrequestv1) to each builder
+via the [`submitBuilderPreferences`][submit-builder-preferences-api] API call in
+the epoch prior to the epoch in which they will be proposing, as determined from
+`state.lookahead`. This ensures builders have the preferences before the bid
```
jtraglia: It should be `state.proposer_lookahead`.

specs/gloas/builder.md
```diff
@@ -0,0 +1,212 @@
...
+`BuilderPreferencesV1` have been submitted, the builder MUST NOT include an
+execution layer payment (i.e. MUST set `bid.execution_payment` to `0`).
+
+*Note*: `bid.value` and `bid.execution_payment` are not mutually exclusive. A
+builder MAY set both fields on a single bid; in that case the builder is
+committing to pay the proposer the sum of the two. `bid.value` is deducted from
+the builder's staked collateral on-chain even when `bid.execution_payment` is
+also set.
+
+## Constructing a `SignedExecutionPayloadEnvelope`
```
jtraglia, line 195: Not sure if you want to specify this (because I do not want to in consensus-specs) but this would be an appropriate place to talk about when a builder should reveal their payload with some certainty that it will be included onchain, within some risk tolerance.

specs/gloas/validator.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### `BuilderPreferencesV1`
+
+`BuilderPreferencesV1` communicates a proposer's per-builder preferences to a
+specific builder ahead of the bid request.
+
+```python
+class BuilderPreferencesV1(Container):
+    max_execution_payment: Gwei
+```
```
jtraglia, line 69: I sort of suggest simplifying this to just be `PreferencesV1`. Because to me it sounds like these are the builder's preferences, not the preferences for each builder. Might just be me.

specs/gloas/builder.md
```diff
@@ -0,0 +1,212 @@
...
+    # Verify parent root
+    # [Modified in Gloas:EIP7732]
+    assert parent_root == hash_tree_root(state.latest_block_header)
+```
+
+## Builder Preferences
+
+Validators MAY communicate their per-builder preferences ahead of the bid
+request by calling the
+[`submitBuilderPreferences`][submit-builder-preferences-api] API in the epoch
```
jtraglia, line 95: I might suggest calling this something like `prepareExecutionPayloadBid` instead. Builders would use this as a signal to start working on a bid. Right now the name doesn't make this exactly clear.

specs/gloas/builder.md
bharath-123: I am unsure if there is value in specifying it. I would assume builders would be sophisticated enough to have their own strategies here. We can revisit this in the future when builders start adapting their infra and there is value in adding it here.

specs/gloas/builder.md
bharath-123: This would be marked with `# [New in Gloas:EIP7732]` cause the check with proposer preferences is pretty new.

specs/gloas/validator.md
bharath-123: Yeah the intention is that, these are the per builder preferences of the validator. I am fine with `PreferencesV1` too, will check what ppl think.

specs/gloas/builder.md
bharath-123: I am not sure if `prepareExecutionPayloadBid` is right here. Builders start working on a bid for proposers ideally every slot based on the [payload attributes SSE event](https://github.com/ethereum/beacon-APIs/blob/master/apis/eventstream/index.yaml#L132).
This endpoint is for validators to be able to submit their preferences. The builder would build their bid based on the latest preference they have seen for the proposer. It is not an indication to start building their bids.

jtraglia: Hmm okay.

apis/builder/execution_payload_bid.yaml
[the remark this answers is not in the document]
rolfyone: If it was configurable there'd want to be strict sanity limits imo, but i think it really does open up some degenerate behaviour to have it available potentially... Probably just not having it settable is the better option than the guards it'd need...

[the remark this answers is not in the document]
tersec: > If it was configurable there'd want to be strict sanity limits imo, but i think it really does open up some degenerate behaviour to have it available potentially... Probably just not having it settable is the better option than the guards it'd need...

Tend to agree with this.

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,177 @@
...
+      required: false
+      description: |
+        Optional header containing the proposer's timeout for the request in milliseconds.
+        Builders should use this header to adjust the amount of time by which they delay the
+        requests to maximise block rewards. Otherwise, requests will timeout and the proposer
+        will not receive the header in time.
+      schema:
+        type: integer
+        format: int64
+        example: 10000
```
rolfyone, line 82: int64 would need to be a string format because of limitations over 54? bits in javascirpt iirc.. @nflaig knows better than me on this though... its the reason most of beacon-api numbers being contained in string anyway...
Not sure if that's a problem here specifically, but figured its worth talking about

bharath-123: The header is sent as ASCII bytes over the wire anyways. I think specifying it as int64 here is good to clarify the type of the header value for the servers parsing the header.

apis/builder/beacon_block.yaml
```diff
@@ -0,0 +1,51 @@
...
+    Submits a `SignedBeaconBlock` to the builder, binding the proposer to the block. 
+
+    A success response (202) indicates that the signed beacon block was
+    valid. If the signed beacon block was invalid, then the builder
+    must return an error response (400) with a description of the validation
+    failure.
+
+    After receiving 202, the proposer takes no further action on the block; the 
+    builder is responsible for publishing the execution payload envelope. The proposer 
+    MAY observe PTC votes to monitor the timely reveal of the execution payload envelope.
```
nflaig: >  The proposer MAY observe PTC votes to monitor the timely reveal of the execution payload envelope.

why is that note here, is there any action you are suggesting the proposer should do, or clients should implement?

apis/builder/builder_preferences.yaml
```diff
@@ -0,0 +1,83 @@
+post:
+  operationId: "submitBuilderPreferences"
+  summary: Submit builder preferences for a proposer.
+  description: |
+    Submits a proposer's `BuilderPreferencesRequestV1` to the builder, including
+    the `max_execution_payment` that the proposer is willing to accept from this
+    builder, authenticated via a `SignedRequestAuthV1`.
+
+    Validators MAY call this endpoint in the epoch prior to the epoch in
+    which they will be proposing, as determined from `state.lookahead`, so that
```
nflaig: this should be `state.proposer_lookahead`

apis/builder/beacon_block.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    This API is applicable from Gloas fork onwards.
+  tags:
+    - Builder
+  parameters:
+    - in: header
+      schema:
+        $ref: "../../builder-oapi.yaml#/components/schemas/ConsensusVersion"
+      required: false
```
nflaig: we should make this required, or how exactly do you determine the consensus type otherwise?

apis/builder/builder_preferences.yaml
```diff
@@ -0,0 +1,83 @@
...
+  parameters:
+    - name: validator_pubkey
+      in: path
+      required: true
+      description: "The BLS public key of the validator expressing these preferences."
+      schema:
+        $ref: "../../beacon-apis/types/primitive.yaml#/Pubkey"
+    - name: Eth-Consensus-Version
+      in: header
+      required: false
```
nflaig: same here, we should make this header required

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,174 @@
...
+    - The hash of the execution layer block the proposer will build on.
+    - The root of the beacon block the proposer will build on.
+    - The index of the proposer.
+    - Optionally, a `SignedRequestAuthV1` in the request body that
+      authenticates the request. The body MAY be encoded as JSON
+      (`Content-Type: application/json`) or SSZ
+      (`Content-Type: application/octet-stream`).
+
+    The proposer's `max_execution_payment` is communicated exclusively via the
+    [`submitBuilderPreferences`][submit-builder-preferences-api] endpoint. If no
```
nflaig: `[submit-builder-preferences-api]` is this a placeholder?

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,174 @@
...
+
+    The proposer sends a POST request to the builder with the following information:
+    - The slot for which the block should be proposed.
+    - The hash of the execution layer block the proposer will build on.
+    - The root of the beacon block the proposer will build on.
+    - The index of the proposer.
+    - Optionally, a `SignedRequestAuthV1` in the request body that
+      authenticates the request. The body MAY be encoded as JSON
+      (`Content-Type: application/json`) or SSZ
+      (`Content-Type: application/octet-stream`).
```
nflaig: it shouldn't be necessary to denote the content-type header here again, it is already in the schema


and we have this general note on the spec

Image

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,174 @@
...
+      required: false
+      description: |
+        Optional header containing the proposer's timeout for the request in milliseconds.
+      schema:
+        type: integer
+        format: int64
+        example: 10000
+    - name: Eth-Consensus-Version
+      in: header
+      required: false
```
nflaig: I noted this before, I think this header should be required

apis/builder/execution_payload_bid.yaml
```diff
@@ -0,0 +1,174 @@
...
+      requested slot. If absent, the builder MAY still serve a bid subject
+      to its own policy.
+    required: false
+    content:
+      application/json:
+        schema:
+          $ref: "../../types/gloas/request_auth.yaml#/Gloas/SignedRequestAuthV1"
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `SignedRequestAuthV1` bytes. Use Content-Type header to indicate that SSZ data is contained in the request body. The `Eth-Consensus-Version` header MUST also be set."
```
nflaig:
```suggestion
          description: "SSZ serialized `SignedRequestAuthV1` bytes. Use Content-Type header to indicate that SSZ data is contained in the request body."
```

we should always sent the version header

builder-oapi.yaml
```diff
@@ -53,10 +53,16 @@ paths:
     $ref: "./apis/builder/validators.yaml"
   /eth/v1/builder/header/{slot}/{parent_hash}/{pubkey}:
     $ref: "./apis/builder/header.yaml"
+  /eth/v1/builder/execution_payload_bid/{slot}/{parent_hash}/{parent_root}/{proposer_pubkey}:
+    $ref: "./apis/builder/execution_payload_bid.yaml"
+  /eth/v1/builder/beacon_block:
```
nflaig: pretty sure plural is more correct here, see previous `/eth/v2/builder/blinded_blocks`

```suggestion
  /eth/v1/builder/beacon_blocks:
```

examples/gloas/signed_request_auth.json
nflaig, line 1: usually we also have a ssz example, see https://github.com/ethereum/builder-specs/tree/main/examples/bellatrix

apis/builder/execution_payload_bid.yaml
nflaig: did anyone ssz encode a `SignedRequestAuthV1` already and made sure it can be decoded by the receiver? I might be wrong, but arbitrary strings like urls seem problematic

apis/builder/execution_payload_bid.yaml
nflaig: this is already part of current `getHeader`

https://github.com/ethereum/builder-specs/blob/1bf8f4910e904e92144f067c6807072f4e09c90d/apis/builder/header.yaml#L46-L57

I don't think it matters as @bharath-123 already noted as this is not transmitted via json

apis/builder/beacon_block.yaml
bharath-123: Nope that can be removed, that was a old line that shouldn't be there! this was back when we were thinking about payload reorg based on PTC. But regardless this shouldn't be there.

apis/builder/execution_payload_bid.yaml
bharath-123: I have a unit test in buildoor on a sort of round trip with request auth here: https://github.com/ethpandaops/buildoor/pull/81/changes#diff-9e15279d13b1890a9f040f159e5d3431a84f9a3406480a4e304b1bd8d75b55e3R58
We basically marshall it into SSZ, unmarshal it, get its hash tree root, sign it and verify the signature

bharath-123: This is very similar to extra data: https://github.com/ethereum/consensus-specs/blob/6ebb2216ca8d7fdac7c108fcfddb74d0e56c42ab/specs/bellatrix/beacon-chain.md#L174 
which is also a `ByteList[MAX_DATA_BYTES]` and can be any arbitrary string.

examples/gloas/signed_request_auth.json
bharath-123: Yes I will add the SSZ examples. I was considering adding them in a follow up PR.

builder-oapi.yaml
bharath-123: Not too opinionated, we only send back 1 beacon block hahaha. But I am fine with this change

apis/builder/execution_payload_bid.yaml
bharath-123: yeah clients should it send it regardless of SSZ or JSON encoding, it seems useful to have in general

apis/builder/beacon_block.yaml
bharath-123: yes agreed, i was not sure if we should make it required because we don't do it in the previous apis. But I see no harm in making it required. A builder can use the header when they need to.

apis/builder/execution_payload_bid.yaml
bharath-123: Shouldn't have added it here. Removing it!

examples/gloas/signed_request_auth.json
bharath-123: Claude was able to get the ssz examples from the json correctly without any issues!

apis/builder/beacon_block.yaml
nflaig: we are doing payload reorgs based on PTC, so builders, if  you read this, watch out

apis/builder/execution_payload_bid.yaml
nflaig: yes please :pray:

builder-oapi.yaml
nflaig: but this is common REST naming convention, we also call it `/eth/v2/beacon/blocks` on the beacon-api

apis/builder/execution_payload_bid.yaml
nflaig: if we are fine with `data` being hex encoded in the json and not human readable, then this works. worth pointing out that depending on what character encoding you use to convert an arbitrary string to bytes it may produce different results, but I am assuming everyone is using `utf-8` and for urls, this is mostly irrelevant even with other encodings like `ascii`

PR #144, opened 2026-01-13
Specify Builder Configurations and Bid selection strategies

bharath-123:
TODO: update PR description


specs/gloas/validator.md
```diff
@@ -233,6 +265,98 @@ block on top of a beacon `state` must take the following actions:
...
+
+```python
+def get_builder_deadline(slot_start_time: uint64) -> uint64:
+    """
+    Calculate deadline for builder responses.
+    Must leave time for local fallback if needed.
+    """
+    # Builders must respond by 2 seconds into the slot
+    # This leaves 2 seconds for local building fallback
+    return slot_start + BUILDER_DEADLINE_MS
```
bharath-123: This deadline can be just 1 sec.

specs/gloas/validator.md
```diff
@@ -233,6 +265,98 @@ block on top of a beacon `state` must take the following actions:
...
+        )
+
+        if signed_bid.message.value < builder_config.min_bid:
+            continue
+        
+        if signed_bid is not None:
+            # Validate the bid
+            registration = registrations.get(builder_config.url)
+            if registration and validate_bid(state, registration, signed_bid, registration.message.fee_recipient):
+                bids.append((builder_config, signed_bid))
```
bharath-123: Bid needs to be boosted before being added.

specs/gloas/validator.md
```diff
@@ -64,6 +74,37 @@ class SignedRequestAuth(Container):
     signature: BLSSignature
 ```
 
+### `BuilderConfig`
+
+```python
+class BuilderConfig(Container):
```
bharath-123: MEV-Boost could be a special case of just having one builder in the configuration pointing to the url of the local mev-boost container.

tersec:
> The main functionality of MEV-Boost is to maintain a whitelist of builders

How/where? https://github.com/flashbots/mev-boost/blob/develop/cli/flags.go selects from entities the current builder API MEV ecyosystem calls relays (which, "whitelist" is an odd word here: it's mostly that they have to be queried explicitly, not that other relays are providing information which is being ignored). It doesn't know or care about what the current builder API MEV ecosystem calls "builders".

Yes, ePBS changes this terminology, but in terms of discussing what `mev-boost` the builder API software does, it's worth being clear about the terminology it uses.

specs/gloas/bid_scoring.md
```diff
@@ -0,0 +1,117 @@
+# Gloas - Bid Scoring
+
+## Overview
+
+This document specifies how a validator selects the best execution payload bid
+from among bids received over the p2p network, bids received via the offchain
```
JasonVranek, line 6: the use of 'offchain' and 'on-chain' is a bit overloaded throughout the docs. 'offchain' refers to both the builder API delivery channel and implicitly the trusted payment, while 'on-chain' refers to the protocol-enforced collateral. Maybe something like 'p2p' vs 'direct' for bid sources and 'trustless' vs 'trusted' for payment types would be clearer?

JasonVranek:
recommend closing as this has been superseded (although very related to this PR's approach!) by a combination of PRs:
- bid scoring has been addressed in [beacon api 630](https://github.com/ethereum/beacon-APIs/pull/630) and builder spec #165
- builder configs has been addressed in [keymanager 88](https://github.com/ethereum/keymanager-APIs/pull/88) and 630 as well

nflaig:
per https://github.com/ethereum/builder-specs/pull/144#issuecomment-5666532232, closing

PR #153, opened 2026-02-23
[validator.md] Fix array

nalepae:
**Before:**
image

**After:**
image


bharath-123:
fixed the root cause for these mis-formatted constants table here: https://github.com/ethereum/builder-specs/pull/154 . Its an issue with the linter. 
Thank you for raising this!

PR #160, opened 2026-05-13
Endpoint to submit builder preferences 

bharath-123:
# Endpoint to submit builder preferences (Gloas)

## Summary

Introduces a new `POST /eth/v1/builder/builder_preferences` endpoint for the
Gloas fork that allows validators to submit per-builder preferences to a builder
ahead of the bid request. This replaces the `X-Eth-Max-Trusted-Bid` request
header as the primary channel for communicating `max_trusted_bid`, while keeping
the header as an optional fallback.

## New types

**`BuilderPreferences`** (`types/gloas/builder_preferences.yaml`)
```python
class BuilderPreferences(Container):
    max_trusted_bid: Gwei
    validator_pubkey: BLSPubkey
```

**`BuilderPreferencesRequest`** — the request body for the new endpoint:
```python
class BuilderPreferencesRequest(Container):
    preferences: BuilderPreferences
    auth: SignedRequestAuth
```

Authentication reuses the existing `SignedRequestAuth` mechanism. The builder
checks that `auth.message.builder_pubkey` matches its own identity and verifies
the BLS signature against `preferences.validator_pubkey`.

## New endpoint

`POST /eth/v1/builder/builder_preferences`

- Validators MAY call this endpoint in the epoch prior to the epoch in which
  they will be proposing, as determined from `state.lookahead`.
- The builder stores the preferences per-validator and uses the `max_trusted_bid`
  when constructing bids.
- `max_trusted_bid` resolution order on `getExecutionPayloadBid`:
  1. Stored `BuilderPreferences` (submitted via this endpoint).
  2. `X-Eth-Max-Trusted-Bid` header, if present and no preferences stored.
  3. `0`, if neither is available.

## Changes to `getExecutionPayloadBid`

- `X-Eth-Max-Trusted-Bid` is now **optional** (`required: false`). Validators
  MAY send it if they have not submitted `BuilderPreferences` to this builder.
- If stored `BuilderPreferences` exist for the proposer, they take precedence
  over the header.
- Removed the `MissingMaxTrustedBid` 400 error example.

## Spec updates

- `specs/gloas/validator.md`: New `BuilderPreferences` and
  `BuilderPreferencesRequest` containers; new _Submitting Builder Preferences_
  section describing when and how to call the endpoint; updated _Bid Request_
  section to reflect the optional header; updated `validate_bid` note.
- `specs/gloas/builder.md`: New _Builder Preferences_ section describing
  storage, precedence rules, and signature verification; updated _Per-request
  Validator Inputs_ to reflect the optional header.

### DIFF
--- a/apis/builder/builder_preferences.yaml
+++ b/apis/builder/builder_preferences.yaml
@@ -2,11 +2,14 @@ post:
   operationId: "submitBuilderPreferences"
   summary: Submit builder preferences for a proposer.
   description: |
-    Submits a proposer's `BuilderPreferences` to the builder, including the
+    Submits a proposer's `SignedBuilderPreferences` to the builder, including the
     `max_trusted_bid` that the proposer is willing to accept from this builder.
 
-    The `builder_pubkey` in the body MUST match the builder's own identity.
+    The `builder_pubkey` in the message MUST match the builder's own identity.
     If it does not, the builder MUST return a 400 response.
+
+    The builder MUST verify the BLS signature against `message.proposer_pubkey`.
+    If signature verification fails, the builder MUST return a 400 response.
 
     A success response (200) indicates that the preferences were accepted. If
     the preferences are invalid, then the builder MUST return an error response
@@ -16,12 +19,12 @@ post:
   tags:
     - Builder
   requestBody:
-    description: A `BuilderPreferences` object expressing the proposer's per-builder preferences.
+    description: A `SignedBuilderPreferences` object expressing the proposer's per-builder preferences.
     required: true
     content:
       application/json:
         schema:
-          $ref: "../../types/gloas/builder_preferences.yaml#/Gloas/BuilderPreferences"
+          $ref: "../../types/gloas/builder_preferences.yaml#/Gloas/SignedBuilderPreferences"
   responses:
     "200":
       description: Success response.
--- a/apis/builder/builder_preferences.yaml
+++ b/apis/builder/builder_preferences.yaml
@@ -1,0 +1,44 @@
+post:
+  operationId: "submitBuilderPreferences"
+  summary: Submit builder preferences for a proposer.
+  description: |
+    Submits a proposer's `BuilderPreferences` to the builder, including the
+    `max_trusted_bid` that the proposer is willing to accept from this builder.
+
+    The `builder_pubkey` in the body MUST match the builder's own identity.
+    If it does not, the builder MUST return a 400 response.
+
+    A success response (200) indicates that the preferences were accepted. If
+    the preferences are invalid, then the builder MUST return an error response
+    (400) with a description of the validation failure.
+
+    This API is applicable from Gloas fork onwards.
+  tags:
+    - Builder
+  requestBody:
+    description: A `BuilderPreferences` object expressing the proposer's per-builder preferences.
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: "../../types/gloas/builder_preferences.yaml#/Gloas/BuilderPreferences"
+  responses:
+    "200":
+      description: Success response.
+    "400":
+      description: Error response.
+      content:
+        application/json:
+          schema:
+            $ref: "../../builder-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            WrongBuilder:
+              value:
+                code: 400
+                message: "builder_pubkey does not match this builder's identity"
+            InvalidPreferences:
+              value:
+                code: 400
+                message: "Invalid builder preferences: max_trusted_bid malformed"
+    "500":
+      $ref: "../../builder-oapi.yaml#/components/responses/InternalError"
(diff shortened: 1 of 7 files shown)


specs/gloas/validator.md
```diff
@@ -89,22 +157,6 @@ The validator then constructs the `SignedRequestAuth` by signing the
 lets builders authenticate the requesting validator and discard requests from
 other parties (e.g. DDOS or replay attempts from competing builders).
 
-### `max_trusted_bid`
```
bharath-123, line 92: I think the definition of `max_trusted_bid` is more suited in `spec/gloas/builder.md` as it is more relevant to builders

apis/builder/builder_preferences.yaml
```diff
@@ -0,0 +1,51 @@
...
+    Validators MAY call this endpoint in the epoch prior to the epoch in
+    which they will be proposing, as determined from `state.lookahead`, so that
+    builders have the preferences before the bid request arrives.
+
+    The builder MUST verify the BLS signature in `auth` against
+    `preferences.validator_pubkey`, and check that `auth.message.builder_pubkey`
+    matches its own identity. If either check fails, the builder MUST return a
+    400 response.
+
+    A success response (200) indicates that the preferences were accepted. If
```
JasonVranek: consider 202

specs/gloas/validator.md
```diff
@@ -55,13 +58,80 @@ class SignedRequestAuth(Container):
...
 
+#### `BuilderPreferences`
+
+`BuilderPreferences` communicates a proposer's per-builder preferences to a
+specific builder ahead of the bid request.
+
+```python
+class BuilderPreferences(Container):
+    max_trusted_bid: Gwei
+    validator_pubkey: BLSPubkey
```
JasonVranek: pubkey could also live in the path like:

`POST /eth/v1/builder/builder_preferences/{validator_pubkey}` which is more consistent with get bid

specs/gloas/builder.md
```diff
@@ -75,22 +76,58 @@ def is_eligible_for_bid(
...
+  - `validator_pubkey`: The BLS public key of the validator submitting these
+    preferences.
+- `auth`: A `SignedRequestAuth` authenticating the request. The builder MUST
+  check that `auth.message.builder_pubkey` matches its own identity and MUST
+  verify the BLS signature against `preferences.validator_pubkey`. If either
+  check fails, the builder MUST return a 400 response.
+
+The builder MUST store the preferences for each proposer and apply the
+`max_trusted_bid` constraint when constructing bids. If no preferences have been
+submitted for a proposer, the builder MUST treat the proposer's `max_trusted_bid`
```
JasonVranek, line 98: MUST seems rather harsh, if it's SHOULD a builder could still return a mixture of bid.value and bid.execution_payment

apis/builder/builder_preferences.yaml
```diff
@@ -0,0 +1,51 @@
...
+    (400) with a description of the validation failure.
+
+    This API is applicable from Gloas fork onwards.
+  tags:
+    - Builder
+  requestBody:
+    description: A `BuilderPreferencesRequest` containing the proposer's preferences and a `SignedRequestAuth` for authentication.
+    required: true
+    content:
+      application/json:
```
JasonVranek, line 36: to be consistent with bids should also support ssz

specs/gloas/validator.md
bharath-123: Yeah we could do that. We use proposer_index for the api to get the bid but I think we can just use the pubkey to keep it backward compaitable with how we do it today

specs/gloas/builder.md
bharath-123: Its definitely not enforceable. I think the right wording would be to say that the builder must respect the `max_trusted_bid` while constructing the bid.

specs/gloas/builder.md
```diff
@@ -75,22 +76,58 @@ def is_eligible_for_bid(
...
+
+Validators MAY communicate their per-builder preferences ahead of the bid
+request by calling the [`submitBuilderPreferences`][submit-builder-preferences-api]
+API in the epoch prior to the epoch in which they will be proposing, as
+determined from `state.lookahead`. The builder receives a `BuilderPreferencesRequest` object containing:
+
+- `preferences`: A `BuilderPreferences` with:
+  - `max_trusted_bid`: The maximum trusted execution layer payment the proposer
+    will accept from this builder (in Gwei).
+  - `validator_pubkey`: The BLS public key of the validator submitting these
```
JasonVranek: since lives in path doesn't need to live in BuilderPreferences anymore

specs/gloas/builder.md
```diff
@@ -75,22 +76,58 @@ def is_eligible_for_bid(
...
+    preferences.
+- `auth`: A `SignedRequestAuth` authenticating the request. The builder MUST
+  check that `auth.message.builder_pubkey` matches its own identity and MUST
+  verify the BLS signature against `preferences.validator_pubkey`. If either
```
JasonVranek:
```suggestion
  verify the BLS signature against `validator_pubkey`. If either
```

PR #161, opened 2026-05-26
Gloas Builder-API spec updates

bharath-123:
The following changes made to the builder-api spec:

1. Rename `max_trusted_bid` to `max_execution_payment`
2. Remove `X-Eth-Max-Trusted-Bid`. The max_execution_payment is only communicated via the builder preferences endpoint. If builder has not seen the max_execution_payment, they can either pay via a trustless payment or choose to not serve the bid.
3. Add a signing domain to request auth
4. Use `proposer_pubkey` instead of `proposer_index` in query params to get the bid and send the builder preferences.
5. Update `is_eligible_for_bid` to check if the validator corresponding to `proposer_pubkey` is active and is the current proposer for the slot.
6. Sign over the builder URL instead of the builder pubkey in the RequestAuth
7. Suffix RequestAuth and BuilderPreferences with V1 in order to version changes.


bharath-123:
Will merge these changes to maintain one single branch

PR #162, opened 2026-06-02
remove submitBeaconBlock API

bharath-123:
When a proposer selects a bid from a builder via the builder-api, only the builder whose bid has been selected can reveal the payload. So the beacon block should be returned to the builder whose bid has been selected. 

This becomes a problem for stateless validator clients as when the validator client sends the beacon node the signed block, just based on the block you cannot tell if the bid is from the builder-api or from the p2p . You need to maintain some state in the beacon node which tells where the bid in the block has come from which makes it trickier for validator clients which want to talk to multiple beacon nodes. This makes the block proposal flow more stateful.

We could also argue that builders could help in improving block dissemnation.

I am on the fence on this and I am waiting for client devs to implement it and get more info on this.


JasonVranek:
I think there's a lot of value in keeping `submitBeaconBlock`. Today highly connected relays assist in broadcasting signed blocks before the attestation deadline which increases the viability/robustness of more geographically isolated proposers. In ePBS, the auction's winning builder would equally be incentivized to help with the propagation. If we remove this endpoint arguably the same geographically isolated proposers would be disadvantaged.

vladimir-ea:
I am sure there is a discussion somewhere which sets out the rationale for this change - could it be referenced here please?

bharath-123:
Apologies, i should have updated the description. The rationale is as follows:
When a proposer selects a bid from a builder via the builder-api, only the builder whose bid has been selected can reveal the payload. So the beacon block should be returned to the builder whose bid has been selected. 

This becomes a problem for stateless validator clients as when the validator client sends the beacon node the signed block, just based on the block you cannot tell if the bid is from the builder-api or from the p2p . You need to maintain some state in the beacon node which tells where the bid in the block has come from which makes it trickier for validator clients which want to talk to multiple beacon nodes. This makes the block proposal flow more stateful

I am on the fence on this and I am waiting for client devs to implement it and get more info on this.

bharath-123:
> I think there's a lot of value in keeping `submitBeaconBlock`. Today highly connected relays assist in broadcasting signed blocks before the attestation deadline which increases the viability/robustness of more geographically isolated proposers. In ePBS, the auction's winning builder would equally be incentivized to help with the propagation. If we remove this endpoint arguably the same geographically isolated proposers would be disadvantaged.

We can also make the argument that the beacon block's size goes down significantly with ePBS and that block broadcast can be done with a much lower bandwidth compared to today. The block could disseminate much faster in the network.

nflaig:
I agree with https://github.com/ethereum/builder-specs/pull/162#issuecomment-4661861631 and in favor of keeping the api

terencechain:
In favor of keeping the api

nflaig:
closing as there is no support for removing this api

PR #165, opened 2026-07-13
Clarifications on builder spec

JasonVranek:
There have been some ambiguities when reading the builder specs around slot numbers, URL encodings, request auth signing/verifying, plus some broken links. This PR aims to clarify them.


shane-moore:
From the SSV side, the proposal-slot clarification and reuse of one auth for the AOT and JIT channels are compatible with our pre-signing design. I think one signing-contract issue remains.

`RequestAuthV1.data` contains bytes, but `canonicalize` operates on URL text. The current code passes the byte-valued `request_auth.data` directly to a string function, which raises a type error. Conversely, the canonical URL returned by that function must be ASCII-encoded before it can populate the SSZ `ByteList`.

More importantly for SSV, normalizing inside verification authenticates a reconstructed message rather than the exact `signed_request_auth.message`. If operators do not derive the same exact message and signing root, their partial signatures cannot be reconstructed.

Could we instead canonicalize the configured builder URL before constructing `RequestAuthV1`, then have the builder compare the received `message.data` with its own canonical URL bytes?

This maps onto the existing spec structure: steps 1 and 2 belong in `specs/gloas/validator.md` under "Constructing the `RequestAuthV1`" and "URL canonicalization"; step 3 belongs in `specs/gloas/builder.md` under "Signing"; and steps 4 and 5 belong in that file's existing request-validation text. `types/gloas/request_auth.yaml` should then mirror the resulting byte-level meaning of `data`.

1. Canonicalize and validate configured URL text.
2. ASCII-encode it and construct `RequestAuthV1(data=canonical_bytes, slot=proposal_slot)`.
3. Sign `compute_signing_root(request_auth, domain)` without transforming it.
4. On the builder, require `message.data` to equal the builder's canonical identity bytes, then verify `compute_signing_root(message, domain)` unchanged.
5. Reject noncanonical wire `data`, even if normalizing it would identify the same builder.

JasonVranek:
5078eabc63cf092007bb933ae16209087c053eba decouples the request authorization data from the builder's URL. The motivation is that request authorizations require data unique to the builder. A builder's URL is just one instance of this, so rather than lock into one format, the idea is that the authorization data is just opaque bytes. These can be pre-negotiated between the builder and proposer, or very well just be the builder's URL. This change just makes it more flexible. 

Related to the conversation here ethereum/beacon-APIs/pull/625

gd-0:
Thanks @JasonVranek !

Very supportive of changes that:
1. include ahead-of-time proposer preferences
2. natively support sidecars and their future extensions

specs/gloas/builder.md
```diff
@@ -9,8 +9,10 @@
   - [Builder Preferences](#builder-preferences)
     - [`max_execution_payment`](#max_execution_payment)
   - [Per-request Validator Inputs](#per-request-validator-inputs)
+    - [Routing through a proxy](#routing-through-a-proxy)
```
potuz: I think all explicit mentions to a proxy should be completely removed from the spec. This is out of protocol and unnecessary to be mentioned. 

We just need to have two fields, one is the `url` to send the request, and another is arbitrary `data`. The latter is part of the signed auth, the former is not. 

These two fields are already enough to enable proxying if people want to use that.

potuz (changes requested):

specs/gloas/builder.md
JasonVranek: Agreed, will remove explicit mentions of proxy across the PRs. The proxy was included because of the change to make auth `data` opaque bytes and not explicitly carry the builder's `url`. For a proxy to exist under this format, they must set `url` to their own and *require* `data` to carry the builder's URL to route the request. An easily fixable issue is that the constraint that per-key, URLs must be unique should be relaxed to (`url`, `data`) pairs should be unique. This would allow for example, a relay to receive `N` requests to different builders that carry real auth `data`. 

For example the config would need to support:
```json
{
      "0xa057816155ad77931185101128655c0191bd0214c201ca48ed887f6c4c6adf334070efcd75140eada5ac83a92506dd7a": {
        "fee_recipient": "0x50155530FCE8a85ec7055A5F8b2bE214B3DaeFd3",
        "target_gas_limit": "45000000",
        "builder": {
          "enabled": true,
          "builders": [
            {
              "url": "https://proxy.example.com",
              "data": "https://builder-a.example.com",
              "pubkey": "0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a",
              "max_execution_payment": "250000000",
              "min_bid": "10000000",
              "builder_boost_factor": "100"
            }
          ],
          "max_execution_payment": "500000000"
        },
        {
        "url": "https://proxy.example.com", // this should be allowed (will update PRs)
        "data": "https://builder-b.example.com",
        ...
        }
      }
```

apis/builder/execution_payload_bid.yaml
```diff
@@ -62,41 +65,35 @@ post:
...
+        plus `X-Timeout-Ms`, and the proposer discards later responses.
       schema:
         type: integer
         format: int64
         example: 10000
-    - name: Eth-Consensus-Version
-      in: header
-      required: true
-      description: |
-        The active consensus version to which the `SignedRequestAuthV1` in the
```
nflaig, line 87: so in the future if we wanna have a `SignedRequestAuthV2` we bump the whole api version? I do agree this isn't fork versioned per se, but the mechanism could still be used as such

I am still fine with bumping the whole api though, just wanna be clear what versioning strategy we wanna use here

apis/builder/execution_payload_bid.yaml
```diff
@@ -62,41 +65,35 @@ post:
         $ref: "../../beacon-apis/types/primitive.yaml#/Pubkey"
     - name: Date-Milliseconds
       in: header
-      required: false
+      required: true
```
nflaig, line 68: why is this change in this PR? it seems already quite convoluted with different changes

apis/builder/builder_preferences.yaml
```diff
@@ -24,18 +30,12 @@ post:
   tags:
     - Builder
   parameters:
-    - name: validator_pubkey
+    - name: proposer_pubkey
```
nflaig, line 33: related to https://github.com/ethereum/beacon-APIs/pull/630#discussion_r3650913143, maybe there was some discussion around this but I don't really understand why we wanna sent a single request per pubkey, maybe someone can explain this

apis/builder/execution_payload_bid.yaml
JasonVranek: Not opinionated on this I just carried over the conventions from Bharath's original version

apis/builder/execution_payload_bid.yaml
JasonVranek: I can put it back but this was a simplification born from having the keymanager api expect fully resolved fields including the request auth -> it's always going to be able to be supplied. The original protest (largely by me) of having required request auths was when it was difficult/unclear how a proxy could get/use them.

apis/builder/builder_preferences.yaml
JasonVranek: I don't think I understood the question the first time. Is the assumption if a node operator has more than one proposer in the lookahead they can save on a call if batching was supported?

nflaig: yes, you only do a single call no matter how many validators are proposers in the lookahead, that matches how the proposer preferences api works

as far as I am aware all apis support batching, this one seems like an exception that decided to do a request per pubkey, but I don't see why it needs to be

to be clear, I don't feel strongly about this, on mainnet that will not make a large difference, but it would be kinda nice if the builder preferences work alongside the proposer preferences more closely unless there is a good reason to have a different api design

apis/builder/execution_payload_bid.yaml
nflaig: I don't understand the relation, making `Date-Milliseconds` header required seems to be completely separate to me, but seems like you referring to the auth param, for that I agree is should be required and probably default to builder url if not supplied

apis/builder/execution_payload_bid.yaml
nflaig: >  the conventions from Bharath's original version

so we bump the api from v1 to v2 if we wanna introduce a `SignedRequestAuthV2`?

apis/builder/execution_payload_bid.yaml
JasonVranek: oops I misread. I was talking about request auths. for the timing headers I folded in based on the convo from discord https://discord.com/channels/595666850260713488/1528987631474446466/1529900444288155709

nflaig: yep I know the discord discussion and agree with making them required in gloas, but a separate PR seems like the better way to go about just to get some approvals on those changes easier, also not quite sure how to merge anything in this repo, there is no maintainer since Bharath is gone

apis/builder/beacon_blocks.yaml
```diff
@@ -20,9 +20,9 @@ post:
     - in: header
       schema:
         $ref: "../../builder-oapi.yaml#/components/schemas/ConsensusVersion"
-      required: true
+      required: false
       name: Eth-Consensus-Version
-      description: "The active consensus version to which the block being submitted belongs."
+      description: "The active consensus version to which the block being submitted belongs. Required if request is SSZ encoded."
```
nflaig: no please, this is wrong, the only reason this was not required on the old apis is becasue ssz was added retrospectively in https://github.com/ethereum/builder-specs/pull/104, please keep them required for new apis, I was telling this @bharath-123 over and over becasue he did the same thing, not having the headers required is just a mess imo

nflaig: I am also curious what would be the rationale for having them optional for json, there isn't much of a benefit, it's nice metadata to have in both cases

nflaig: see this issue why it's a mess, relying on a note/comment instead of the schema is bad, for example in lodestar we run spec tests to make sure all headers are set, this helps avoid such issues
- https://github.com/ethereum/builder-specs/pull/118

JasonVranek: will revert it 🫡

0w3n-d (approved):

apis/builder/builder_preferences.yaml
nflaig: after thinking about this more and reading up on the current spec, although I haven't found that mentioned as an explicit argument during the initial spec design, I believe it makes sense due to signed auth requests since we allow pubkeys to sign over different auth, so there can be partial failures, however per the spec [here](https://github.com/ethereum/builder-specs/blob/main/apis/builder/builder_preferences.yaml) this should return a 401 which is communicated back via http status code which doesn't really allow reporting partial failures. It seems favorable to send individual requests per-pubkey to me.

specs/gloas/builder.md
```diff
@@ -94,24 +95,29 @@ Validators MAY communicate their per-builder preferences ahead of the bid
...
+  signature against the `proposer_pubkey` path parameter and MUST check that
+  `auth.message.data` matches the value it agreed with the proposer, so that an
+  unauthenticated or replayed request cannot skew a proposer's preferences away
+  from the value the proposer chose. If the signature fails to verify, the
+  builder MUST return a 401 response; if the `auth.message.data` check fails,
+  the builder MUST return a 400 response. The builder MUST also reject, with a
+  400 response, preferences whose `auth.message.slot` has already passed, so
+  that a replayed request cannot roll preferences back to a stale value.
+
+The builder SHOULD store the preferences for each proposer and MUST honor the
```
nflaig: so should this be stored in a tuple such as `(proposer, slot)`, it's not very likely but possible that a validator has two proposals in the lookahead

JasonVranek: more specific in 9a58e07

examples/gloas/builder_preferences_request.json
eserilev, line 14: nit: this still lists preferences before auth. This PR swaps the order of the fields. Doesn't matter for JSON but it is important for SSZ. Keeping the order consistent would be nice from a clarity POV

types/gloas/builder_preferences.yaml
```diff
@@ -1,18 +1,18 @@
...
       max_execution_payment:
         $ref: "../../beacon-apis/types/primitive.yaml#/Uint64"
-        description: "Indicates the maximum amount that a proposer is willing to accept as an execution payment from the builder."
-  BuilderPreferencesRequestV1:
+        description: "The maximum amount, in Gwei, that a proposer is willing to accept as an execution payment from the builder."
+  BuilderPreferencesRequest:
     type: object
     description: "A builder preferences submission containing the proposer's preferences and a signed request authentication."
-    required: [preferences, auth]
+    required: [auth, preferences]
```
eserilev: alternatively to https://github.com/ethereum/builder-specs/pull/165/changes#r3781619840 we  just dont switch the ordering of the fields here. I'm not sure why the order was switched, but if theres no reason for it I would actually lean towards not swapping the order of these fields

JasonVranek: I thought there was a request to reorder but it might have been in one of the other PRs. I'm unopinionated, do people prefer swapping the order back or fixing the order in the JSON example?

nflaig: no opinion on this one, but I do remember I left a comment about ordering here https://github.com/ethereum/beacon-APIs/pull/630#discussion_r3698487106

> but if theres no reason for it I would actually lean towards not swapping the order of these fields

seems like a good argument to keep the original ordering if it's already implemented in eg. buildoor

JasonVranek: ah okay i'll return it to `[preferences, auth]`

specs/gloas/builder.md
```diff
@@ -94,24 +95,29 @@ Validators MAY communicate their per-builder preferences ahead of the bid
...
+  signature against the `proposer_pubkey` path parameter and MUST check that
+  `auth.message.data` matches the value it agreed with the proposer, so that an
+  unauthenticated or replayed request cannot skew a proposer's preferences away
+  from the value the proposer chose. If the signature fails to verify, the
+  builder MUST return a 401 response; if the `auth.message.data` check fails,
+  the builder MUST return a 400 response. The builder MUST also reject, with a
+  400 response, preferences whose `auth.message.slot` has already passed, so
+  that a replayed request cannot roll preferences back to a stale value.
+
+The builder SHOULD store the preferences per proposer per `auth.message.slot`
```
nflaig, line 117: might not be worth to point this out explicitly but I suspect that only the latest valid submission will be considered? builders can just handle it the same way they do for `registerValidator` today. I could definitely see that validator client re-config and restart could trigger such updates in edge cases

JasonVranek: `max_execution_payment` is not signed over, I believe this was so that `SignedRequestAuth` could apply cleanly to both `getExecutionPayloadBid` and `submitBuilderPreferences` and avoid adding another signed type. So if a MITM accessed the `SignedRequestAuth` they could replay and change a proposer's `max_execution_payment`.

nflaig: > max_execution_payment is not signed over

ah right, so builders wanna use the first value they see, I suppose there isn't much reason for a proposer to change that value within the lookahead

> MITM accessed the SignedRequestAuth they could replay and change a proposer's max_execution_payment

the beacon node needs to re-verify the bid for sure, so worst case the builder loses the bid if this happens as the bn rejects it

apis/builder/builder_preferences.yaml
```diff
@@ -2,19 +2,25 @@ post:
...
     builders have the preferences before the bid request arrives.
 
-    The builder MUST verify the BLS signature in `auth` against
-    `validator_pubkey`, and check that `auth.message.data`
-    matches its own URL. If the signature check fails, the builder MUST
-    return a 401 response. If the URL or slot check fails, the builder MUST
-    return a 400 response.
+    The body MAY be encoded as JSON (`Content-Type: application/json`) or SSZ
+    (`Content-Type: application/octet-stream`); `BuilderPreferencesRequest`
+    is fork-versioned, so the `Eth-Consensus-Version` header is required.
```
nflaig: I find this note a bit unncessary, the schema already defines this without room for interpretation
- `Gloas/BuilderPreferencesRequest` (fork versioned)
- `Eth-Consensus-Version` has `required: true`
- `content` defines both `application/json`  and `application/octet-stream`

apis/builder/builder_preferences.yaml
```diff
@@ -62,25 +68,27 @@ post:
             SlotMismatch:
               value:
                 code: 400
-                message: "Invalid SignedRequestAuthV1: auth.message.slot does not match the requested slot"
-            WrongBuilder:
+                message: "Invalid SignedRequestAuth: auth.message.slot has already passed"
```
nflaig: I guess another reason for `SlotMismatch`, or actually the more accurate reason would be if the validator is not proposer of the given slot?

also compared to `ProposerPreferences`, the `BuilderPreferencesRequest` doesn't contain the `dependent_root` so in case the builder and proposer have a different view of the chain, it's not trivial for them to validate it

but this might be fine for this message since it's not propagated via gossip and builders can decide to keep `BuilderPreferencesRequest` even if it doesn't match their current view of the shuffling

apis/builder/execution_payload_bid.yaml
```diff
@@ -5,32 +5,35 @@ post:
...
+    The proposer's beacon node sends a POST request to the builder with the following information:
     - The slot for which the block should be proposed.
     - The hash of the execution layer block the proposer will build on.
     - The root of the beacon block the proposer will build on.
-    - The index of the proposer.
-    - Optionally, a `SignedRequestAuthV1` in the request body that
-      authenticates the request. The body MAY be encoded as JSON or SSZ.
+    - The public key of the proposer.
+    - A `SignedRequestAuth` in the request body that authenticates the
+      request. The body MAY be encoded as JSON or SSZ.
```
nflaig:
```suggestion
      request.
```

apis/builder/execution_payload_bid.yaml
```diff
@@ -5,32 +5,35 @@ post:
...
-    If it is absent, the builder MAY still serve a bid, but builders MAY
-    use the presence and validity of the `SignedRequestAuthV1` to apply
-    per-validator policy (e.g. rate-limiting, prioritization, or refusing
-    unauthenticated requests).
+    The `SignedRequestAuth` body is required. If the body is missing or
+    malformed, the request is invalid and the builder MUST return a 400
+    response. If its `auth.message.data` or `auth.message.slot` does not match,
+    the builder MUST return a 400 response; if its signature fails to verify,
+    the builder MUST return a 401 response. Builders MAY use the authenticated
+    identity to apply per-validator policy (e.g. rate-limiting).
```
nflaig: to be clear on this, so the CL client is expected to always send this, but a builder may choose to ignore it, right?

specs/gloas/builder.md
```diff
@@ -129,33 +135,37 @@ reputation.
...
+- A required `Date-Milliseconds` header with the Unix timestamp in milliseconds
+  at which the request was sent, and a required `X-Timeout-Ms` header with the
+  proposer's timeout for the request, measured from `Date-Milliseconds`. The
+  timestamp lets the builder estimate the transit delay from proposer to
+  builder; the builder MUST respond by `Date-Milliseconds` plus `X-Timeout-Ms`,
+  and the proposer discards later responses.
+
+The `Eth-Consensus-Version` header is required on every request that carries a
+body, naming the fork of the body's type. The builder MUST return a 400 response
+when the header is absent or names a fork it does not recognize.
```
nflaig: this whole section seems redundant to me, but fine with keeping it in case someone only looks at the `.md` file(s)

specs/gloas/builder.md
```diff
@@ -205,12 +251,13 @@ The specification for a block builder to construct a
...
 documented in the [Gloas consensus specs][gloas-builder-specs].
 
+[bls]: https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/beacon-chain.md#bls-signatures
 [get-execution-payload-bid-api]: ./../../apis/builder/execution_payload_bid.yaml
 [gloas-builder-specs]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/builder.md
 [gloas-consensus-specs]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas
-[proposer-preferences]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md
-[proposer-preferences-topic]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md
+[proposer-preferences]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#new-proposerpreferences
+[proposer-preferences-topic]: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#new-proposer_preferences
```
nflaig, line 264: all of these links should probably be fixed later by using a tag/release instead of master as reference

types/gloas/request_auth.yaml
```diff
@@ -1,22 +1,30 @@
...
     type: object
     required: [data, slot]
     properties:
       data:
         type: string
         format: hex
-        description: "Opaque authentication data, up to `MAX_DATA_SIZE` bytes. Typically set to the URL of the builder the request is intended for."
-        example: "0x68747470733a2f2f6275696c6465722e6578616d706c652e636f6d"
-        pattern: "^0x[a-fA-F0-9]{0,8192}$"
+        description: >-
```
nflaig, line 9: I suppose this is the same as `|`? never seen it used before

nflaig (commented):
generally looks good to me, just some comments and few self-reminders

apis/builder/builder_preferences.yaml
JasonVranek: dropped in ce6251d

james-prysm (approved):

specs/gloas/validator.md
```diff
@@ -39,73 +39,76 @@ corresponding to the included bid to the PTC committee.
...
 
 ```python
-class RequestAuthV1(Container):
+class RequestAuth(Container):
     data: ByteList[MAX_DATA_SIZE]
     slot: Slot
 ```
 
-#### `SignedRequestAuthV1`
+#### `SignedRequestAuth`
```
nflaig: since we are renaming these, I think it would be good to use less generic names, what about the following

- `RequestAuth` / `SignedRequestAuth` -> `BuilderRequestAuth` / `SignedBuilderRequestAuth`
- `DOMAIN_REQUEST_AUTH` -> `DOMAIN_BUILDER_REQUEST_AUTH`
- `MAX_DATA_SIZE` -> `MAX_BUILDER_AUTH_DATA_SIZE`
- `get_/verify_request_auth_signature` -> `get_/verify_builder_request_auth_signature`

when reading the builder-specs itself, the naming is fine, but in client code it's a bit too generic

cc @markolazic01 who brought this up

apis/builder/execution_payload_bid.yaml
```diff
@@ -5,32 +5,37 @@ post:
...
 
     The proposer's `max_execution_payment` is communicated exclusively via the
-    `submitBuilderPreferences` endpoint. If no `BuilderPreferencesV1` have been 
-    submitted for the proposer, the builder MUST treat `max_execution_payment` as `0` 
-    and MUST NOT include an execution layer payment in the bid.
+    `submitBuilderPreferences` endpoint. Any bid the builder serves MUST honor
+    the `max_execution_payment` cap from stored preferences. Without stored
+    preferences the builder MAY serve a bid with any `execution_payment`; the
+    proposer's locally configured per-builder limits are the backstop: the
+    proposer values a bid at its `value` plus `min(execution_payment,
```
nflaig, line 23: in `validate_bid` we still have this `assert bid.execution_payment <= max_execution_payment`

and in `validator.md` is still says

> A builder MUST honor the `max_execution_payment` cap in any bid it serves for a
slot it has stored preferences for; without them it MAY serve a bid with any
`execution_payment`. The proposer's locally configured per-builder limits are
the backstop: the proposer discards any bid that exceeds them.

(same note is in `builder.md`)

and another note in `builder.md`

>  Without
stored preferences the builder MAY set any `bid.execution_payment`; the proposer discards any bid that exceeds its locally configured limits.

should probably adjust all of these too

wordlist.txt
```diff
@@ -35,8 +35,9 @@ Gloas
 gloas
 Gwei
 PTC
-BuilderPreferencesV
-RequestAuthV
-SignedRequestAuthV
+UTF
+BuilderPreferences
+RequestAuth
+SignedRequestAuth
```
nflaig: can be cleaned up
```suggestion
UTF
```

specs/gloas/validator.md
```diff
@@ -231,28 +239,29 @@ block on top of a beacon `state` must take the following actions:
...
    [Gloas validator specs][gloas-validator-specs] but with the best
    [`SignedExecutionPayloadBid`][signed-execution-payload-bid] from the prior
    step.
-3. The proposer returns the `SignedBeaconBlock` back to the upstream block
-   building software via [`submitSignedBeaconBlock`][submit-signed-beacon-block]
-   API call.
+3. The validator publishes the `SignedBeaconBlock` to its beacon node, which
+   returns it to the upstream block building software via
+   [`submitSignedBeaconBlock`][submit-signed-beacon-block] API call. The
+   validator does not call builders directly.
```
nflaig, line 251: > The validator does not call builders directly.

I don't know why we have such details here in the spec, in case of vero/vouch they might directly talk to builders, in any case, this note isn't harmful, I am mostly questioning this whole section here, not just this specific line

specs/gloas/validator.md
```diff
@@ -181,7 +192,6 @@ are also defined in the consensus specs.
 def validate_bid(
     state: BeaconState,
     proposer_preferences: ProposerPreferences,
-    max_execution_payment: uint64,
     signed_bid: SignedExecutionPayloadBid,
     fee_recipient: ExecutionAddress,
 ) -> bool:
```
nflaig, line 197: `min_bid` is missing here, might be something we can add in a follow-up if desired, or it's fine as is since it's a local selection rule, and not related to bid validity

markolazic01:
LGTM, thanks for the renames @nflaig @JasonVranek

nflaig (approved):

PR #168, opened 2026-09-16
Derive default auth data from the builder hostname

nflaig:
The default `data` is the UTF-8 bytes of the builder URL exactly as advertised, so a trailing `/`, an explicit `:443` or different casing on the proposer side fails the builder's check.

Use the builder's hostname instead: lowercased, without scheme, userinfo, port or path. It is a pure function of the configured URL, so all validator clients derive the same bytes, and it is only applied when constructing the `BuilderRequestAuth`, the signed bytes are still verified unchanged.

Builders that need a finer identity than their hostname can still agree `data` out of band. The `auth_data` descriptions in keymanager-APIs and beacon-APIs are updated in https://github.com/ethereum/keymanager-APIs/pull/94 and https://github.com/ethereum/beacon-APIs/pull/642.


specs/gloas/validator.md
```diff
@@ -159,6 +160,38 @@ The validator then constructs the `SignedBuilderRequestAuth` by signing the
 validator and discard requests from other parties (e.g. DDOS or replay attempts
 from competing builders).
 
+#### Default auth data
+
+The default `data` for a builder URL is the builder's hostname: the lowercased
+ASCII hostname of the URL, an IPv6 literal written in its compressed form
+([RFC 5952][rfc-5952]) inside brackets. Scheme, userinfo, port, path, query and
```
JasonVranek:
```suggestion
The default `data` is the hostname of the builder's URL: 
lowercased, in its ASCII form, with an IPv6 literal 
written in its compressed form ([RFC 5952][rfc-5952])
inside brackets.
```

0w3n-d (approved):
Looks good to me. I'm a fan of defaults like this.

JasonVranek (approved):

syjn99 (approved):

StefanBratanov (approved):

ckoopmann (approved):
Yes this would work for the ultrasound relay as well.

